PR #32, opened 2020-03-21
Ben

bentheiii:
various optimizations and code cleanups


bentheiii:
only the last commit actually matters

PR #73, opened 2020-03-24
Agent states

bentheiii:
This is a massive update to the source code that reduces the entire medical state to a finite-state machine, making it easy to add/change states down the line. It includes many changes that needs to be looked at.
It also includes numerous optimizations and code tweaks


bentheiii:
I'm closing this for now, will re-open soon

PR #84, opened 2020-03-26
Tal

tal-nc:
fixed a small bug in markovic matrix generation. 
changed average_infection_length to average time in each state, returning a dict of state to how much time on average you spend in this state.
improved & fixed normalization (r0 is now high again)

### DIFF
--- a/src/corona_hakab_model/consts.py
+++ b/src/corona_hakab_model/consts.py
@@ -13,7 +13,7 @@ from util import dist, upper_bound
 
 class Consts(NamedTuple):
     # simulation parameters
-    population_size = 10_000
+    population_size = 100_000
     total_steps = 400
     initial_infected_count = 20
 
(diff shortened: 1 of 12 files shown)


src/corona_hakab_model/consts.py
```diff
@@ -1,18 +1,18 @@
...
 from scipy.stats import rv_discrete
 from state_machine import StochasticState, TerminalState
-from util import dist
+from util import dist, upper_bound
 
 
 class Consts(NamedTuple):
     # simulation parameters
-    population_size = 10_000
+    population_size = 100_000
```
mor-m-shemesh: This changes the memory usage dramatically.
Is there a reason to change the default to 100k?

tal-nc: not really, everyone can feel free to play with this number.
i ran the tests before uploading this on 100k because this allows you to see the results of the normalization changes more precisely.

omri312: I agree that this should remain 10_000 in our branch, so it will be faster to test yourself and see results.
You'll run 100_000 when you need to check that.

PR #87, opened 2020-03-27
Supervisor upgrade

tal-nc:
added a static method to supervisor, to compare between different simulations runs.
added optional pre-fix and style parameters to all supervisables classes.
left an example in main of how to use the newly added comparison option.
on the way, deleted a few unused methods.

(this branch contains the changes from branch tal, which upgrades the normalization function, ignore them on this branch)

### DIFF
--- a/src/corona_hakab_model/supervisor.py
+++ b/src/corona_hakab_model/supervisor.py
@@ -83,17 +83,17 @@ class Supervisor:
             )
         if self.manager.consts.home_isolation_sicks:
             title = (
-                title
-                + "\napplying home isolation for confirmed cases ({} of cases)".format(
-                    self.manager.consts.caught_sicks_ratio
-                )
+                    title
+                    + "\napplying home isolation for confirmed cases ({} of cases)".format(
+                self.manager.consts.caught_sicks_ratio
+            )
             )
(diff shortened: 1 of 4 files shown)


src/corona_hakab_model/supervisor.py
```diff
@@ -130,7 +168,7 @@ def scale(self) -> Optional[Tuple[float, float]]:
         pass
 
     @abstractmethod
-    def plot(self, ax):
+    def plot(self, ax, prefix="", style=""):
```
bentheiii: I'm not sure about this, Maybe it's better to override name and style with a special class and not allow the supervisor free reign. We should talk about this change

src/corona_hakab_model/supervisor.py
```diff
@@ -119,6 +119,44 @@ def plot(self, max_scale=True, auto_show=True, save=True):
         if auto_show:
             plt.show()
 
+    @staticmethod
+    def static_plot(simulations_info: Sequence[("SimulationManager", str, Sequence[str])], title="comparing",
```
bentheiii: If we want to merge plots, we need some better architecture, not something ad-hoc like this

PR #92, opened 2020-03-27
Export data to a file

omri312:
New PR to develop

### DIFF
--- a/src/corona_hakab_model/manager.py
+++ b/src/corona_hakab_model/manager.py
@@ -2,6 +2,7 @@ import logging
 import random as rnd
 from itertools import islice
 from time import time
+import ast 
 
 import infection
 import numpy as np
@@ -10,8 +11,6 @@ import update_matrix
 from affinity_matrix import AffinityMatrix
 from consts import Consts
 from medical_state import MedicalState
-import json
-import csv
 
 class SimulationManager:
     """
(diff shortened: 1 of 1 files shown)


omri312:
It seems that this was implemented in the manager, but should now move to the supervisor, then this can be merged

src/corona_hakab_model/manager.py
```diff
@@ -131,7 +131,26 @@ def run(self):
...
+
+    def Export(self):
+        consts_attr = [attr for attr in dir(self.consts) if not callable(getattr(self.consts, attr)) and not attr.startswith("_")]
+        output_dict = {}
+        for attr in consts_attr:
+            output_dict[attr] = getattr(self.consts,attr)
+        for attr in self.per_generation:
+            output_dict[str(attr)] = self.per_generation[attr]
+        
+        filehandle = open("..\output\Export.json","w")
```
mor-m-shemesh, line 144: Why have it be a static file path? just get it as an argument

src/corona_hakab_model/manager.py
```diff
@@ -131,7 +131,26 @@ def run(self):
...
+        export_string = "{\n"
+        for key, val in output_dict.items():
+            export_string = export_string + '"' + str(key) + '": ' + str(val) + ',\n'
```
mor-m-shemesh, line 147: Why not use json.dump?
It seems you are already exporting only simple data types

src/corona_hakab_model/manager.py
mor-m-shemesh: Also, always use 
with open(<file path>, 'wb') as f:
    ....
so you don't mess with handles to files.

src/corona_hakab_model/manager.py
mishana, line 147: an f-string will look much better here

omri312:
I believe this is irrelevant, closing.

PR #93, opened 2020-03-28
Add json

TsvikaD:
Now all the constants are comming from a json file (currently hard coded the file name within the manager)


src/corona_hakab_model/consts.py
```diff
@@ -3,121 +3,119 @@
...
+        We sanitize the loaded data, for obvious reasons.
+        Documentation about what is allowed and not allowed can be found at the top of this page.
+        """
+        with open(param_path, "r") as read_file:
+            data = read_file.read()
 
-    def set_distribution_params(self, distributions_dict):
-        valid_distribution_keys = self.simulation_parameters["distributions"].keys()
+        Consts.sanitize_data(data)
+        parameters = eval(data)
```
bentheiii: remove builtins and add dist to the eval namespace

bentheiii: how does this even run?

src/corona_hakab_model/consts.py
```diff
@@ -3,121 +3,119 @@
...
+    "family_strength": 0.4,
+    "work_strength": 0.04,
+    "stranger_strength": 0.004,
+}
+
+ConstParameters = namedtuple('ConstParameters', sorted(default_parameters))
+
+
+class Consts(ConstParameters):
+
```
bentheiii: give this class a __slots__?

src/corona_hakab_model/consts.py
```diff
@@ -3,121 +3,119 @@
...
-                raise UnknownParameterException(key)
+        We sanitize the loaded data, for obvious reasons.
+        Documentation about what is allowed and not allowed can be found at the top of this page.
+        """
+        with open(param_path, "r") as read_file:
+            data = read_file.read()
 
-    def set_distribution_params(self, distributions_dict):
-        valid_distribution_keys = self.simulation_parameters["distributions"].keys()
+        Consts.sanitize_data(data)
```
bentheiii: as far as I can tell this function doesn't exist

src/corona_hakab_model/consts.py
mor-m-shemesh: Not possible, as it is already a subclass of a namedtuple class.

src/corona_hakab_model/consts.py
mor-m-shemesh: It inherits the current scope's globals and locals.
Fixed, I have passed specific functions to the eval, such as dist and rv_discrete

src/corona_hakab_model/consts.py
mor-m-shemesh: 6 lines down :)

src/corona_hakab_model/consts.py
bentheiii: it is possible, and as it currently stands the class is unslotted

bentheiii: a subclass of namedtuple can have empty slots, and if not, the class is unslotted

src/corona_hakab_model_data/default_params.json
```diff
@@ -0,0 +1,38 @@
...
+    "stop_work_days": 30,
+    "resume_work_days": 60,
+    "average_family_size": 5,
+    "average_work_size": 50,
+    "average_amount_of_strangers": 200,
+    "family_strength_not_workers": 0.75,
+    "family_strength": 0.4,
+    "work_strength": 0.04,
+    "stranger_strength": 0.004
+}
```
bentheiii: this file should be an empty dict

src/corona_hakab_model_data/default_params.json
```diff
@@ -0,0 +1,38 @@
...
+        "silent_to_asymptomatic_days": "(0, 3, 10)",
+        "silent_to_symptomatic_days": "(0, 3, 10)",
+        "asymptomatic_to_recovered_days": "(3, 5, 7)",
+        "symptomatic_to_asymptomatic_days": "(7, 10, 14)",
+        "symptomatic_to_hospitalized_days": "(0, 1.5, 10)",
+        "hospitalized_to_asymptomatic_days": "(18)",
+        "hospitalized_to_icu_days": "(5)",
+        "icu_to_deceased_days": "(7)",
+        "icu_to_hospitalized_days": "(7)"
+    },
```
bentheiii: these should be flattened (no sub-dicts)

bentheiii: also calling dist

src/corona_hakab_model/consts.py
```diff
@@ -1,38 +1,119 @@
...
+        Load parameters from file and return Consts object with those values.
+
+        No need to sanitize the eval'd data as we disabled __builtins__ and only passed specific functions
+        Documentation about what is allowed and not allowed can be found at the top of this page.
+        """
+        with open(param_path, "rt") as read_file:
+            data = read_file.read()
+
+        parameters = eval(data, {"__builtins__": None, "dist": dist, "rv_discrete": rv_discrete})
+
```
bentheiii: If the researches send us some unhashable data types (lists or dicts), the program will crash in expected places, maybe we should check here that the return value is hashable?

src/corona_hakab_model_data/default_params.json
mor-m-shemesh: This file is deprecated, I deleted it

src/corona_hakab_model/consts.py
mor-m-shemesh: Will add back "sanitize_parameters" to check exactly that

src/corona_hakab_model_data/default_params.json
mor-m-shemesh: Done.
Added appropriate documentation

src/corona_hakab_model/consts.py
mor-m-shemesh: Done

src/corona_hakab_model/consts.py
```diff
@@ -112,9 +112,15 @@ def from_file(param_path):
...
 
         parameters = eval(data, {"__builtins__": None, "dist": dist, "rv_discrete": rv_discrete})
+        Consts.sanitize_parameters(parameters)
 
         return Consts(**parameters)
 
+    @staticmethod
+    def sanitize_parameters(parameters):
+        for key, value in parameters.items():
+            assert isinstance(value, Hashable), f"{key} of value {value} in parameter file is unhashable"
```
bentheiii: a user could enter as a value the tuple (12,{}), which is an instance of Hashable, but is still not hashable

The best solution is to try to hash all of `parameters` (no need to go element-by element) and fail if the hashing fails

bentheiii: also, assert should only be used for sanity checking, not data validation

mor-m-shemesh: Fixed - now throws InvalidParameterException

PR #94, opened 2020-03-28
Matrix generation

tal-nc:
changed the matrix generation technique.
now you can easily choose which connections you want to model, and control their distribution.
the generation is heavy and should be under a lot of profiling (even tho most of the time goes on filling the lil matrices themselves).

also removed a few un-used functions and attributes.

i tried to do this as efficient as i could, but would appreciate your optimizations proposals.

### DIFF
--- a/src/corona_hakab_model/affinity_matrix.py
+++ b/src/corona_hakab_model/affinity_matrix.py
@@ -88,6 +88,9 @@ class AffinityMatrix:
                 total_contagious_probability += time_in_state * state.contagiousness
             beta = self.consts.r0 / total_contagious_probability
 
+            # saves this for the effective r0 graph
+            self.total_contagious_probability = total_contagious_probability
+
             # this factor should be calculated once when the matrix is full, and be left un-changed for the rest of the run.
             self.factor = (beta * self.size) / (self.matrix.sum())
 
--- a/src/corona_hakab_model/affinity_matrix.py
+++ b/src/corona_hakab_model/affinity_matrix.py
@@ -46,19 +46,21 @@ class AffinityMatrix:
         self.agents = self.manager.agents
 
         self.logger.info("Building circular connections matrices")
-        self.circular_matrices = [
+        # all circular matrixes. keeping as a tuple of matrix and type (i.e, home, work, school and so)
+        self.circular_matrices = [(
             self.circular_matrix_generation(self.agents, matrix.circle_size_probability,
-                                            matrix.connection_strength).tocsr() for
+                                            matrix.connection_strength).tocsr(), matrix.type)for
             matrix in self.consts.circular_matrices]
 
         self.logger.info("Building non circular connections matrices")
-        self.non_circular_matrices = [
-            self.non_circular_matrix_generation(self.agents, matrix.scale_factor, matrix.connection_strength).tocsr()
+        # all non-circular matrixes. keeping as a tuple of matrix and type (i.e, home, work, school and so)
+        self.non_circular_matrices = [(
+            self.non_circular_matrix_generation(self.agents, matrix.scale_factor, matrix.connection_strength).tocsr(), matrix.type)
             for matrix
             in self.consts.non_circular_matrices]
 
         self.logger.info("summing all matrices")
-        self.matrix = sum(self.circular_matrices) + sum(self.non_circular_matrices)
+        self.matrix = sum(matrix[0] for matrix in self.circular_matrices) + sum(matrix[0] for matrix in self.non_circular_matrices)
 
         self.factor = None
         self.normalize()
@@ -96,6 +98,14 @@ class AffinityMatrix:
         # switching from probability to ln(1-p):
         non_zero_keys = self.matrix.nonzero()
         self.matrix[non_zero_keys] = np.log(1 - self.matrix[non_zero_keys])
+
+    def change_connections_policy(self, types_of_connections_to_use: List[str]):
+        self.logger.info(f"changing policy. keeping all matrixes of types: {types_of_connections_to_use}")
+        self.matrix = sum(matrix[0] for matrix in self.circular_matrices if matrix[1] in types_of_connections_to_use)\
+                      + sum(matrix[0] for matrix in self.non_circular_matrices if matrix[1] in types_of_connections_to_use)
+        self.normalize()
+
+
 
     def non_circular_matrix_generation(self, agents_to_use, scale_factor: float, connection_strength):
         """
--- a/src/corona_hakab_model/affinity_matrix.py
+++ b/src/corona_hakab_model/affinity_matrix.py
@@ -1,9 +1,13 @@
 import logging
 from random import shuffle
-
 import numpy as np
-from agent import TrackingCircle
+from agent import TrackingCircle, Agent
 from scipy.sparse import lil_matrix
+from util import dist
+from typing import List, Dict
+from scipy.stats import rv_discrete
+import math
+from sub_matrices import CircularConnectionsMatrix, NonCircularConnectionMatrix
 
 m_type = lil_matrix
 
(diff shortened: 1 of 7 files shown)


src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+        :param circle_size_probability: representing the size probability
+        :param connection_strength: the wanted connection strength
+        :return: lil_matrix with the wanted connections
+        """
+        my_agents_length = len(agents)
+        matrix = m_type((self.size, self.size), dtype=np.float32)
+
+        # pre-rolling all the sized of the circles for efficiency causes.
+        circles_size_rolls = circle_size_probability.rvs(
+            size=math.ceil(my_agents_length / circle_size_probability.mean()))
```
bentheiii: would should take at least one SD more the we would expect, random number generation is faster in bulk. (circle_size_probability also has a variance).

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+        pre_rolled_circles_len = len(circles_size_rolls)
```
bentheiii: again, this should be removed

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+
+        circles_counter = 0
+        agents_counter = 0
```
bentheiii: should be renamed

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+
+    def circular_matrix_generation(self, agents, circle_size_probability: rv_discrete, connection_strength):
+        """
+        this method will create a matrix of circular connections given a list of agents, an rv_discrete representing the size of the circle probabilty, and the connection strength
+        :param agents: list of all agents
+        :param circle_size_probability: representing the size probability
+        :param connection_strength: the wanted connection strength
+        :return: lil_matrix with the wanted connections
+        """
+        my_agents_length = len(agents)
```
bentheiii: should be renamed

src/corona_hakab_model/affinity_matrix.py
bentheiii: also this should be an iterator

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+                current_circle_size = circle_size_probability.rvs()
+
+            # adding agents to the current circle
+            for _ in range(current_circle_size):
+                if agents_counter >= my_agents_length:
+                    break
+                choosen_agent = agents[agents_ids[agents_counter]]
+                current_circle.add_agent(choosen_agent)
+                # todo possibly add this circle to the agent circles list
+                agents_counter += 1
```
bentheiii: why not pop from agent_ids and spare the trouble?

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+        pre_rolled_circles_len = len(circles_size_rolls)
+
+        circles_counter = 0
+        agents_counter = 0
+
+        # here all the circles will be saved
+        circles: List[TrackingCircle] = []
+
+        agents_ids = np.array([a.index for a in agents])
+        np.random.shuffle(agents_ids)
```
bentheiii: why is this an array and not a list (of agents)?

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+
+        for circle in circles:
+            ids = np.array([a.index for a in circle.agents])
+            xs, ys = np.meshgrid(ids, ids)
+            xs = xs.reshape(-1)
+            ys = ys.reshape(-1)
+            matrix[xs, ys] = connection_strength
+
+        ids = np.arange(my_agents_length)
+        matrix[ids, ids] = 0
```
bentheiii: add a comment on what this does doing this

src/corona_hakab_model/manager.py
```diff
@@ -113,42 +113,18 @@ def setup_sick(self):
...
         runs full simulation
         """
-        self.generate_policy(1)
+        self.setup_sick()
 
         for i in range(self.consts.total_steps):
             if Consts.active_isolation:
                 if i == self.consts.stop_work_days:
-                    self.matrix.change_work_policy(False)
+                    self.matrix.change_connections_policy(["home", "strangers"])
```
bentheiii: use sets instead

src/corona_hakab_model/manager.py
```diff
@@ -113,42 +113,18 @@ def setup_sick(self):
...
                 elif i == self.consts.resume_work_days:
-                    self.matrix.change_work_policy(True)
+                    self.matrix.change_connections_policy(["home", "strangers", "school", "work"])
```
bentheiii: use sets instead

src/corona_hakab_model/sub_matrices.py
```diff
@@ -0,0 +1,26 @@
...
+    __slots__ = "type", "agents", "circle_size_probability", "connection_strength"
+
+    # todo switch type to a named-tuple
+    def __init__(self, type: str, agents: List[Agent], circle_size_probability: rv_discrete, connection_strength: float):
+        self.type = type
+        self.agents = agents
+        self.circle_size_probability = circle_size_probability
+        self. connection_strength = connection_strength
+
+class NonCircularConnectionMatrix:
```
bentheiii: maybe they should have a common super-type?

src/corona_hakab_model/supervisor.py
```diff
@@ -269,3 +275,21 @@ def get(self, manager) -> float:
...
     def name(self) -> str:
         return "Total(" + ", ".join(n.name() for n in self.inners)
+
+class _EffectiveR0Supervisable (FloatSupervisable):
+    def __init__(self):
+        super().__init__()
+
+    def get(self, manager) -> float:
+        # note that this calculation is VARY heavy (that's why it is only calculating once every 5 days)
+        suseptable_indexes = [index for index, val in enumerate(manager.susceptible_vector) if val]
```
bentheiii: use numpy.argwhere (or even numpy.flatnonzero)

src/corona_hakab_model/supervisor.py
```diff
@@ -269,3 +275,21 @@ def get(self, manager) -> float:
...
+        return np.sum(1 - np.exp(item) for item in manager.matrix.matrix[suseptable_indexes, :].data) * manager.matrix.total_contagious_probability / manager.matrix.size
```
bentheiii: calling np.sum on a generator destroys all of np's power
`np.sum(1 - np.exp(manager.matrix.matrix[suseptable_indexes].data))`?

src/corona_hakab_model/supervisor.py
```diff
@@ -269,3 +275,21 @@ def get(self, manager) -> float:
...
+
+    def snapshot(self, manager):
+        if manager.current_date % 5 != 0 and not manager.current_date <= 1:
+            return
```
bentheiii: snapsot shouldn't be overriden, make this check in get() (and return numpy.nan)

src/corona_hakab_model/sub_matrices.py
tal-nc: possible, but i don't think it really matters, at least not for the upcomming version

src/corona_hakab_model/supervisor.py
tal-nc: done. improved by a lot

src/corona_hakab_model/supervisor.py
tal-nc: done

src/corona_hakab_model/supervisor.py
tal-nc: returning numpy.nan destroys the plot (it doesnt know where to place those points so doesnt print at all).
for now i just removed this part, since with the generator removal, it runs fast enough to calculate at every turn.
however, i don't really see any problem with overriding the snapshot.
another possible solution is to update the base snapshot so that it won't add any point that is np.nan

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+        """
+        # the lil_matrix to be returned
+        matrix = m_type((self.size, self.size), dtype=np.float32)
+
+        # calculates it only once for effifiency
+        amount_of_agents = len(agents_to_use)
+
+        # an iterator representing the rolled amount of connections per agent
+        # todo note that alpha is beeing reduced by 0.5, because later on it is getting math.ceil. it will not change the mean but will change the distribution
+        amount_of_connections = map(math.ceil, np.random.exponential(scale_factor - 0.5, amount_of_agents))
```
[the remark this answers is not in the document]
bentheiii: then use np.cast, mapping it like this ruins all of numpy's speed

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+        # dict of agents to amount of remaining connections to make for this agent
+        remaining_contacts: Dict[Agent, int] = {agent: amount for (agent, amount) in
+                                                zip(agents_to_use, amount_of_connections)}
+
+        # will be used as a stopping sign
+        connections_sum = sum(remaining_contacts.values())
+
+        # a list of indexes of agents which still lack connections. used for efficiency
+        available_agents = [agent for agent in agents_to_use]
+        available_agents_len = len(available_agents)
```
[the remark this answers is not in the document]
bentheiii: https://stackoverflow.com/questions/35189516/why-is-lena-list-object-so-slow
len is the same speed, this isn't c++ and calling len isn't slowing your code

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -212,17 +88,169 @@ def normalize(self):
...
+                connections_tuples[0, connections_cnt] = second_agent.index
+                connections_tuples[1, connections_cnt] = current_agent.index
+                connections_cnt += 1
+
+                # note that there is no reason to update firs's remaining connections for efficiency
+                remaining_contacts[second_agent] -= 1
+                connections_sum -= 2
+
+                if remaining_contacts[second_agent] <= 0:
+                    del (available_agents[roll_index])
```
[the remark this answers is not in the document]
bentheiii: deleting an item from the middle of an array is incredibly heavy, and requires O(N) operations

src/corona_hakab_model/affinity_matrix.py
bentheiii: If you only roll `my_agents_length / circle_size_probability.mean()` values, you will have roll more values half the time, if you do `my_agents_length / circle_size_probability.mean() + 10`, then that's less likely. Of course with some match and the SD of the distribution we can get a more accurate additional rolls that will only require re-rolling by some small probability.

src/corona_hakab_model/affinity_matrix.py
bentheiii: len simply retrieves a stored value from the list

src/corona_hakab_model/supervisor.py
bentheiii: Delay already adds NANs, you can do it, it's fine

src/corona_hakab_model/affinity_matrix.py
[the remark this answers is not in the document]
tal-nc: done

src/corona_hakab_model/affinity_matrix.py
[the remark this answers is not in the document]
tal-nc: ok. removed available_agents_len

src/corona_hakab_model/affinity_matrix.py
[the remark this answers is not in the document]
tal-nc: we must keep track on who still lack connections.
the only other way i see is choosing a random agent from the dict at each iterate, which means activating random each loop, and deleting items straight from the dictionary.
do you think it will be better?

src/corona_hakab_model/affinity_matrix.py
tal-nc: ok, switched to len

PR #103, opened 2020-03-30
Matrix generation improvments

tal-nc:
added a new matrix generation method, which creates a clustered connections matrix (meaning it has communities and stimulate societies better).
algo is based on the article: 
https://www.researchgate.net/publication/267872239_Triadic_closure_as_a_basic_generating_mechanism_of_communities_in_complex_networks
a point worth to consider is that the amount of connections distribution, while close to being exponential distribution, ain't completely is.
would appreciate optimization offers, since this is a heavy method (similar complexity to the current non-circular method).
1 last point, i tried to run tox -e quality but it kept getting an invocation error, and file not found error. asked on slack and didn't get any response. i did tox -e reformat so i hope it is ok.


src/corona_hakab_model/affinity_matrix.py
```diff
@@ -33,67 +33,42 @@ def __init__(self, agents: Iterable[Agent], consts: Consts):
...
         for i, cm in enumerate(self.consts.circular_matrices):
             self.circular_matrix_types[cm.name] = (i, cm)
+
         self.non_circular_matrix_types = {}
         for j, ncm in enumerate(self.consts.non_circular_matrices, len(self.circular_matrix_types)):
             self.non_circular_matrix_types[ncm.name] = (j, ncm)
-        self.depth = j + 1
+
+        self.clustered_matrix_types = {}
+        for l, clm in enumerate(self.consts.clustered_matrices,
```
bentheiii, line 46: PEP 8: no variables named 'l'

src/corona_hakab_model/affinity_matrix.py
```diff
@@ -196,27 +168,23 @@ def clustered_matrix_generation(
...
-                [node.index for node in other_nodes], int
-            )
-            connections_cnt += len(other_nodes)
+            # add the newly made connections to the connections list. note that this is one directional,
+            # but the other direction will be added when adding the other's connections
+            connections[nodes[i].index].extend([node.index for node in other_nodes])
 
         # add the rest of the nodes, one at a time
-        for node in nodes[m + 1 :]:
+        for node in nodes[m + 1:]:
```
bentheiii, line 181: slicing copies the list, use itertools.islice instead

tal-nc: done

src/corona_hakab_model/affinity_matrix.py
tal-nc: it was the same way before, just with j... it works...
if you prefer it this way, i declared l before the loop

bentheiii: no no, by PEP 8 conventions, you shouldn't name a variable "l"

https://www.python.org/dev/peps/pep-0008/#names-to-avoid

tal-nc: oh ok. makes more seance.
renamed to k.

PR #104, opened 2020-03-31
Matrix impl

bentheiii:
a massive change that introduces the new kind of sparse matrix: the parasymbolic matrix


bentheiii:
merging with Omer's approval

PR #105, opened 2020-04-01
CoronaMatrix class import selector

idanmiara:


src/corona_hakab_model/affinity_matrix.py
```diff
@@ -40,6 +35,8 @@ def __init__(self, manager, input_matrix_path: str = None, output_matrix_path: s
         for j, ncm in enumerate(self.consts.non_circular_matrices, len(self.circular_matrix_types)):
             self.non_circular_matrix_types[ncm.name] = (j, ncm)
         self.depth = j + 1
+
+        CoronaMatrix = corona_matrix.get_corona_matrix_class(self.manager.consts.use_parasymbolic_matrix)
```
mor-m-shemesh, line 39: Save the type of the matrix as a member and use it when creating new matrices, you may want to created (and should create) extra matrices in other methods of the class.
This will also be useful when others are trying to stay compatible with the AffinityMatrix's types, such as the simulation manager

bentheiii: users can just use type(self.inner)

idanmiara:
will reopen a new PR

PR #110, opened 2020-04-05
Generation remake matrix creation other

rlfreiman:
added generation for community connections


src/corona_hakab_model/generation/matrix_consts.py
```diff
@@ -23,5 +23,7 @@ class MatrixConsts(NamedTuple):
     }
     use_parasimbolic_matrix = True,
     clustering_switching_point = 20,
+    community_member_edges = 2, #m
```
tal-nc, line 26: the community member edges (m) is half the average amount of connections per node. 
in this case, the average amount of connections per node, is given by the daily + weekly amount of connections, for the specific connection type.
since m will be a float, when adding the nodes, some will add math.ceil(m) connections, and some math.floor(m) connections.
u can use the random_round() method, located at the bottom of matrix_generator.py, to choose who will be rounded up and who will be rounded down.

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -162,6 +163,69 @@ def create_random_clustered_circles_matrix(self, con_type: ConnectionTypes, circ
...
+        weekly_connections_float = self.matrix_consts.weekly_connections_amount_by_connection_type[con_type]
+        total_connections_float = daily_connections_float + weekly_connections_float
+        
+        for circle in circles:
+            agents = circle.agents
+            indexes = [agent.index for agent in agents]
+            nodes: List[Node] = [Node(index) for index in indexes]
+            
+            
+            num_agent_edges = self.matrix_consts.community_member_edges[0]
```
tal-nc, line 182: as stated above, should me calculated using total_connections_float (which is the sum of daily and weekly connections).

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -162,6 +163,69 @@ def create_random_clustered_circles_matrix(self, con_type: ConnectionTypes, circ
                                           p=[daily_connections_float / total_connections_float, weekly_connections_float / total_connections_float])
             v = np.full_like(conns, strengthes, dtype=np.float32)
             self.matrix[depth, agent.index, conns] = v
+            
+    def _create_geographically_clustered_circles_matrix(self, con_type: ConnectionTypes, circles: List[SocialCircle], depth):
```
tal-nc, line 167: there was no reference to the geographic circles.
for each node, there should be a higher chance (given as an input) to select a node from your own geographic circle.
this way, you would get a connections graph clustered by geographic circles.

src/corona_hakab_model/generation/matrix_consts.py
rlfreiman: fixed

src/corona_hakab_model/generation/matrix_generator.py
rlfreiman: fixed

src/corona_hakab_model/generation/matrix_generator.py
rlfreiman: renamed

PR #111, opened 2020-04-05
Generation remake matrix creation

tal-nc:
added matrix creation.
added pickle for population data.
missing 'other' connection type matrix creation


src/corona_hakab_model/generation/circles_generator.py
```diff
@@ -0,0 +1,131 @@
...
+        # export population data using pickle
+        with open(self.EXPORT_OUTPUT_DIR + self.EXPORT_FILE_NAME, 'wb') as export_file:
+            pickle.dump(self.population_data, export_file)
+
+    def import_population_data(self, import_file_path=None):
+        if import_file_path is None:
+            import_file_path = self.EXPORT_OUTPUT_DIR + self.EXPORT_FILE_NAME
+
+        with open(import_file_path, 'rb') as import_file:
+            self.population_data = pickle.load(import_file)
```
omri312, line 131: Notice this import won't work, because you don't use ```self.population_data``` at all - 
you should do something similar to the export function and re-assign all the relevant members.

tal-nc: good point. fixed

PR #112, opened 2020-04-05
Simulation Test

michaeleh:


setup.py
```diff
@@ -12,12 +12,11 @@
...
     package_dir={"": "src"},
     packages=["corona_hakab_model", "corona_hakab_model_data"],
-    install_requires=["numpy", "scipy", "scikit-learn", "matplotlib"],
+    install_requires=["numpy", "scipy", "scikit-learn","matplotlib", "pytest"],
     extras_require={
         "quality": ["pytest", "tox", "flake8", "isort", "black"],
         "profiling": ["yappi"],
-        "pretty graphs": ["pyside2"],
-        "building": ["swimport"],
+        "pretty graphs": ["pyside2", "matplotlib"],
```
meirhalachmi, line 19: matplotlib is already required above, we don't need it as an extra too

setup.py
```diff
@@ -12,12 +12,11 @@
     description="Corona Hakab Model Simulator",
     package_dir={"": "src"},
     packages=["corona_hakab_model", "corona_hakab_model_data"],
-    install_requires=["numpy", "scipy", "scikit-learn", "matplotlib"],
+    install_requires=["numpy", "scipy", "scikit-learn","matplotlib", "pytest"],
     extras_require={
         "quality": ["pytest", "tox", "flake8", "isort", "black"],
         "profiling": ["yappi"],
-        "pretty graphs": ["pyside2"],
-        "building": ["swimport"],
```
meirhalachmi, line 20: why did you delete swimport?

tests/simulation_sanity_test/simulation_test.py
```diff
@@ -0,0 +1,27 @@
...
+f = open('simulation_test_ranges.json', 'r')
+range_per_day_dict = json.load(f)
+
+
+def test_no_policy_simulation():
+    """
+    run the simulation with no policy and fixed R0
+    """
+    warnings.warn(UserWarning("Ranges for each state was generate from real sim runs, might be flaky."))
+    consts = Consts(active_isolation=False, population_size=1_000, r0=2.4)
```
meirhalachmi, line 17: There are many default consts that might change and mess with the results. 
Maybe you should export the consts to an external file and load it in the test

setup.py
michaeleh: done

setup.py
michaeleh: done

tests/simulation_sanity_test/simulation_test.py
michaeleh: done

PR #118, opened 2020-04-05
Generation remake

tal-nc:
new generation mechanism.
removed affinity matrix.

generation manager now creates the population, social circles, and matrix (currently only supporting parasymbolic matrix)
simulation manager gets data holders (that will later be serialized and un-serialized), and extract the needed data from them.

most modules got changed a bit to fit the new mechanism


omri312:
You have multiple ".pickle" binary files that are commited, delete them

PR #119, opened 2020-04-05
Mmmc

bentheiii:
added the new mixed-model multi-circle sparse matrix classes.
It isn't mixed in code yet, this PR is for the sparse matrix classes only


src/corona_hakab_model/clustered_matrix/clustered.py
```diff
@@ -0,0 +1,170 @@
...
+
+import numpy as np
+
+# todo test
+# todo benchmark
+from consts import generator
+from sparse_base import SparseBase, ManifestBase
+
+
+class Cluster:
```
mor-m-shemesh, line 12: Should you make the class slotted?

src/corona_hakab_model/clustered_matrix/clustered.py
```diff
@@ -0,0 +1,170 @@
...
+        return ret
+
+    def __getitem__(self, item):
+        p = self.probs[item]
+        if p == 0:
+            return None
+        return p, self.vals[item]
+
+
+class ClusteredSparseMatrix(SparseBase):
```
mor-m-shemesh, line 75: ditto about slotting

src/corona_hakab_model/sparse_matrix/sparse.hpp
```diff
@@ -0,0 +1,68 @@
...
+    public:
+        //invariant: if w_val = 0 or v_val = 0, operate(w_val, v_val) = 1
+        virtual dtype operate(dtype w_val, dtype v_val) const;
+};
+
+class ManifestMatrix;
+
+class SparseMatrix{
+    private:
+        // note, each column set only stores indices that might not be zero, the only guarantee
```
mor-m-shemesh, line 20: this is relevant to the entice cpp codebase, but why not use ptr objects? unique_ptr and whared ptr?
this feels like an opportunity for memory leakage

bentheiii: Swig's support for smart pointers is limited, There aren't very many allocations throughout the code (aside from ctors and dtors)

src/corona_hakab_model/clustered_matrix/clustered.py
bentheiii: done

src/corona_hakab_model/clustered_matrix/clustered.py
bentheiii: done

src/corona_hakab_model/util.py
```diff
@@ -34,6 +36,18 @@ def lower_bound(d):
     return d.a + d.kwds.get("loc", 0)
 
 
+def is_strict_sorted(s: Iterable):
```
mor-m-shemesh, line 39: Iterable not defined - import it

src/corona_hakab_model/clustered_matrix/clustered.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def __init__(self, original: ClusteredSparseMatrix):
+        self.original = original
+        self.inners = [np.zeros_like(c.vals) for c in self.original.clusters]
+
+    def I_POA(self, v: np.ndarray, magic):
+        ret = np.empty(self.original.size, dtype=np.float32)
+        for cluster, manifest in zip(self.original.clusters, self.inners):
+            zipped_v = v[cluster.indices]
+            ret[cluster.indices] = [prod(magic.operate(m, v) for (m, v) in zip(row, zipped_v)) for row in manifest]
```
tal-nc, line 161: i think you should do 1 - prod(...).
you can only prod the probabilities to NOT get infected in each of the meetings (since proding the probabilities of getting infected in each of the meeting would give you the probability you were infected a lot of times).
this way the prod would give you the probability that you weren't infected in any of the meetings, and 1 - prod is the probability that you were infected (at least in one meeting).

you should also make sure magic operator gives you the probability of NOT beeing infected in a single meeting, or otherwise is should be: 1 - prod(1 - magice.operate(m,v))

this applies for the cpp code as well.

PR #120, opened 2020-04-05
initial commit

rlfreiman:
initial code as a basis for discussion


src/corona_hakab_model/update_matrix.py
```diff
@@ -54,9 +74,25 @@ def change_connections_policy(self, connection_types_to_use: Iterable[Connection
...
         self.normalize()
 
-    def update_matrix_step(self):
+    def update_matrix_step(self, circle_policies : Iterable[PolicyByCircles]):
         """
         Update the matrix step
         """
         # for now, we will not update the matrix at all
-        pass
+        for policy_by_circle in circle_policies:
```
omri312, line 82: Like I wrote in slack,
I'd move this implementation to a seperate function in this class called "apply_circle_policies" or so,
As there will be more stages in the updating of the matrix, and this function will eventually call each of them one by one...

src/corona_hakab_model/update_matrix.py
```diff
@@ -1,8 +1,28 @@
 from generation.connection_types import ConnectionTypes
+from generation.circles import SocialCircle
 from agent import Agent
 import numpy as np
-from typing import Iterable
+from typing import Iterable, Callable, Any
 
+class Policy:
```
omri312, line 7: In total looks good and generic,
We need to make sure this also goes for all kinds of policies applied, for example -
- We send only people marked "detected" home, making them unconnected to all society, except family.
- Make some changes only for certain ages of the population

etc...

PR #127, opened 2020-04-06
added policies manager. cleaned code

tal-nc:
added new policies manager that controlls policies changes.
allowed more flex-able policies, and updates to the matrix.
cleaned up the code a bit, and removed some un-used things


src/corona_hakab_model/consts.py
```diff
@@ -102,9 +71,42 @@
         lambda agent: agent.medical_state.name == "Recovered",
         lambda agent: agent.medical_state.name == "Symptomatic",
     ),  # TODO: Define better API
-    # dictionary of {date : percent} that controls what percentage of schools are open
-    "school_openage_factors": {1: 0, 30: 0.5, 45: 1},
-    "should_change_school_openage": False
+
+    # policies data
+    # a dictionary of type day:List[ConnectionTypes]. on each day, keeps only the given connection types opened
+    "policies_changes": {
```
bentheiii, line 77: no dicts in const! use a tuple of tuples instead

src/corona_hakab_model/consts.py
```diff
@@ -102,9 +71,42 @@
...
+    },
+    "change_policies": False,
+
+    # policies acting on a specific connection type, when a term is satisfied
+    "partial_opening_active": True,
+    # each connection type gets a list of conditioned policies
+    # each conditioned policy actives a specific policy when a condition is satisfied
+    # each policy changes the multiplication factor of a specific circle
+    # each policy is activated only if a list of terms is full-filled.
+    "connection_type_to_conditioned_policy": {
```
bentheiii, line 90: no dicts

src/corona_hakab_model/consts.py
```diff
@@ -102,9 +71,42 @@
...
+        ConnectionTypes.School:
+        [
+            ConditionedPolicy(
+                activating_condition=lambda kwargs: len(np.flatnonzero(kwargs["manager"].contagiousness_vector)) > 1000,
+                policy=Policy(0, [lambda circle: random() > 0]), message="closing all schools"),
```
bentheiii, line 95: a. random() > 0 always returns true
b. calling random() for each circle is slow

src/corona_hakab_model/consts.py
```diff
@@ -102,9 +71,42 @@
...
+            ConditionedPolicy(
+                activating_condition=lambda kwargs: len(np.flatnonzero(kwargs["manager"].contagiousness_vector)) < 500,
```
bentheiii, line 97: so many lambdas in a single line is a code smell, use inheritance instead

src/corona_hakab_model/consts.py
tal-nc: why? it would make things a lot uglier.
if this is un-avoidable, is there a not-ugly way too check item existence in a tuple of tuples?
(i would like to know if a given date is in this data structure, not by looping through each item)

src/corona_hakab_model/consts.py
tal-nc: i know, this is only there as a prove of concept that you can close a given percent of the circles at a given connection type (this was the requirement of integration team).
the random() call only occurs if the first condition is true, and the policy is not already active, making it occur only on rare occasions, so it is ok time-wise to roll for every circle.

src/corona_hakab_model/consts.py
tal-nc: how will inheritance fix it?
the lambdas gives the researchers complete freedom without touching anything but consts, and without creating a hard-coded object / function for every different option.

src/corona_hakab_model/consts.py
mor-m-shemesh: What I would do is export these parameters to another place, since they are not Constant values ever present in the simulation, but parameters for the simulation.

The medical state parameters are going to be detached from the consts soon enough too

tal-nc: good idea, i agree.
i think there should be a consts file for each different aspect of the simulation (medical consts, policy consts, plotting, tests, population and etc...)
but just for now, what is the problem with putting a dict in consts? i would rather not switch it to tuples before the version release.

src/corona_hakab_model/consts.py
tal-nc: the researchers in this case are shahf and shir, and they both know how to code.
the contagiousness count in this case is only an example. there are many different possible activation conditions that they wanted to try out (for example, a given date, amount of agents in a given state, and so on).
i don't really see a way doing it without hard-coding an object for any possible condition.
any way, i agree that this whole module should be upgraded, and changed in the next version, but i think this is ok for now

src/corona_hakab_model/agent.py
```diff
@@ -81,17 +81,20 @@ def remove_agent(self, agent):
...
         self.agents = set()
+        # done to count how many different agents ever visited a given state
+        self.ever_visited = set()
 
     def add_agent(self, agent):
         super().add_agent(agent)
         if agent in self.agents:
             raise ValueError("DuplicateAgent")
         self.agents.add(agent)
+        self.ever_visited.add(agent.index)
```
bentheiii, line 97: why store indices instead of agents?

tal-nc: because we only care about the len of it, and ints take less memory than references in python (i think)

bentheiii: all values in python require the same length. an "agent" is just an address pointer

tal-nc: changed to store agents

src/corona_hakab_model/state_machine.py
```diff
@@ -366,7 +366,7 @@ def markovian(self):
 
         # todo remove?
         for col in range(next_index):
-            assert np.sum(ret[:, col]) == 1
+            assert np.sum(ret[:, col]) >= 0.99
```
bentheiii, line 369: we should use `math.isclose` instead

tal-nc: switched to math.isclose

PR #128, opened 2020-04-06
Feature/detection module

meirhalachmi:
Some improvements after feedback


src/corona_hakab_model/healthcare.py
```diff
@@ -36,31 +36,51 @@ def test(self, agent: Agent):
 
 
 class HealthcareManager:
-    def __init__(self, sim_manager: SimulationManager):
+    def __init__(
+            self, sim_manager: SimulationManager,
+            daily_num_of_test_schedule: Dict,
```
tal-nc, line 41: why not extracting this data from manager.consts?

meirhalachmi: No real good reason. It felt kinda wrong to propagate constants all the way to the healthcare manager, but I already did it with the testing gaps.

Anyway, I'll change it as you suggested

PR #131, opened 2020-04-07
consts and matrix_consts - NamedTuple -> dataclass

idanmiara:
why do we have such a complicated mecanisem for Consts?
should we use dataclass (with or without __slots__) ?
I've tested the default run: with NamedTuple it runs as fast as dataclass with __slots__. 
without __slots__ it's 1% slower.
@bentheiii - regarding your comment about no dicts in consts - where would you like to use lru_cache here and if we use dataclass does it still apply?


bentheiii:
NamedTuples are a. faster, b. immutable, c. slot automatically
dataclasses, for instance, are unhashable
I agree that the current situation is a mess but data classes won't improve on it in any way that namedtuples don't. I don't see any reason why namedtuple should be replaced with data class

idanmiara:
> i see that you have deleted consts, but i don't see anything to replace it.
> is it a mistake, or am i missing something?

oops... now it's back

idanmiara:
> NamedTuples are a. faster, b. immutable, c. slot automatically
> dataclasses, for instance, are unhashable
> I agree that the current situation is a mess but data classes won't improve on it in any way that namedtuples don't. I don't see any reason why namedtuple should be replaced with data class

as far as I can see they are comparable regarding speed, but dataclasses are much more readable.
why do you need it to be immutable or hashable?
take a look at the new consts.py

mor-m-shemesh:
The matrix_consts and circles_consts have been updated to fit the reworked consts.py.
I think this PR can be closed :)

PR #145, opened 2020-04-09
added option to import circles consts from file

RoyIsland:


src/corona_hakab_model/generation/circles_consts.py
```diff
@@ -60,11 +81,11 @@ def geographic_circles(self):
             for i in range(self.geo_circles_amount)
         ]
 
-    @cached_property
+    @property
```
mor-m-shemesh, line 84: These are values that won't change.
just make them members in an __init__ function.
Yeah, that's not how it it right now in Consts but it should be

src/corona_hakab_model/generation/circles_consts.py
```diff
@@ -60,11 +81,11 @@ def geographic_circles(self):
...
     def age_distribution(self):
         return rv_discrete(10, 70, values=(self.ages, self.age_prob))
 
-    @cached_property
+    @property
```
mor-m-shemesh, line 88: Ditto of age_distribution

PR #150, opened 2020-04-10
Sparse export

bentheiii:
allow to export sparse matrices


src/corona_hakab_model/main.py
```diff
@@ -54,6 +45,11 @@ def main():
 
     gm = GenerationManger(circles_consts=circles_consts, matrix_consts=matrix_consts)
 
+    if args.sub_command == 'generate':
+        with open(args.output, 'wb') as w:
+            write(gm.matrix_data.matrix, w)
```
mor-m-shemesh, line 50: The matrix_data is not the only thing to save, there are also the agents  and circles.
but as a placeholder i believe it's fine

PR #158, opened 2020-04-10
Expanded state supervisables - current / total / new

meirhalachmi:
This PR includes Freiman's dump-to-CSV PR


src/corona_hakab_model/consts.py
```diff
@@ -29,7 +29,7 @@ class Consts(NamedTuple):
 
     # attributes and default values:
 
-    total_steps: int = 350
+    total_steps: int = 40
```
tal-nc, line 32: switch this back to a higher number before merging

src/corona_hakab_model/agent.py
```diff
@@ -59,21 +59,27 @@ def get_infection_ratio(self):
 
 
 class Circle:
-    __slots__ = "kind", "agent_count"
+    __slots__ = "kind", "agent_count", "added_total", "removed_total"
```
tal-nc, line 62: tracking circle has ever_visited, which is the same as added total. removed total is just ever_visited - agent_count

src/corona_hakab_model/consts.py
meirhalachmi: oops :laughing:

src/corona_hakab_model/agent.py
meirhalachmi: Missed `ever_visited`, and you're right about the `removed_total`. I'll change it

src/corona_hakab_model/supervisor.py
```diff
@@ -34,129 +34,41 @@ def snapshot(self, manager):
...
+            published_data = [v.publish() for v in self.supervisables]
+            csv_writer = csv.writer(output_file)
+            # write data rows
+            for i in range(self.manager.consts.total_steps + 1):
+                row = [i]
+                if i == 0:  # header row
+                    row = ["days"]
+                for data in published_data:
+                    # first column is always the #day
+                    row = row + data[i][1:]
```
mishana, line 51: isn't this gonna attach real data to the header row as well?

tal-nc: i think that the first row of each supervisable is its name, which should be in the header row

meirhalachmi: Anyway I'm switching most of this code to use Pandas

PR #172, opened 2020-04-11
Fix C++ for parasymbolic so that it compiles

benglewis:


src/corona_hakab_model/parasymbolic_matrix/parasymbolic.cpp
```diff
@@ -135,7 +135,7 @@ std::vector<std::vector<size_t>> CoffedSparseMatrix::non_zero_columns(){
         auto& row = rows[row_num];
         std::vector<size_t> el;
         el.reserve(row.size());
-        for (auto& it = row.cbegin(); it != row.cend(); it++){
+        for (auto it = row.cbegin(); it != row.cend(); it++){
```
mishana, line 138: better use auto&&

bentheiii: agreed

benglewis: OK, does someone want to explain that to me?

benglewis: @mishana Thank you for the explanation :)

PR #174, opened 2020-04-11
Feature/issue 163 csv combiner

bgmoshe:


src/corona_hakab_model/analyzers/simulation_analysis.py
```diff
@@ -0,0 +1,24 @@
+import os
+from glob import glob
+from project_structure import OUTPUT_FOLDER
+import pandas as pd
+from datetime import datetime
+
+
+def create_comparison_files(parameters_to_compare: list, files: list = None):
```
meirhalachmi, line 8: Use the headers from the CSV as columns

bgmoshe: How is the new solution? Using set and columns I get all the columns (Notice that this IS affected by casing, but I am not sure if I should deal with it for now, I cna check)

src/corona_hakab_model/analyzers/simulation_analysis.py
```diff
@@ -0,0 +1,34 @@
...
+        result_df = pd.DataFrame(columns=dfs_dict.keys())
+        for file_name, file_df in dfs_dict.items():
+            if parameter in file_df.columns:
+                result_df.loc[:, file_name] = file_df.loc[:, parameter]
+            else:
+                result_df.loc[:, file_name] = pd.nan
+        result_df.to_csv(os.path.join(result_folder_name, f"{parameter}.csv"))
+
+
+create_comparison_files()
```
meirhalachmi, line 34: This line will make the function run anytime this module is imported. remove this line \ move it to a __main__ function.

meirhalachmi: ץ

bgmoshe: DONE

PR #177, opened 2020-04-11
added layers

bentheiii:


src/corona_hakab_model/encounter_layer.py
```diff
@@ -0,0 +1,29 @@
...
+    def add_layer(self, layer: EncounterLayer):
+        self.layers[layer.name] = layer
+
+    def POA(self, v):
+        manifests = {
+            n: m.matrix.manifest() for (n, m) in self.layers.items()
+        }
+        m_iter = iter(manifests.items())
+        first_label, first_manifest = next(m_iter)
+        c = first_manifest.I_POA(v, self.layers[first_label].magic_op)
```
tal-nc, line 26: why excluding the first manifest? you can just announce c = 1 before the loop

bentheiii: The first manifest is used to create the initial C, all the rest are added to it

src/corona_hakab_model/encounter_layer.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict
+
+from sparse_base import SparseBase
+
+
+class EncounterLayer:
+    def __init__(self, name: str, magic_op, matrix: SparseBase):
```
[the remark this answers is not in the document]
bentheiii: done

src/corona_hakab_model/encounter_layer.py
tal-nc: you can just initiate c to 1, but nvm, won't stop the merge for that

PR #190, opened 2020-04-12
Feature/run generation from folders

mor-m-shemesh:
This is a template - 
there will need to be a complementing PR that implements GenerationManager.save_to_folder

### DIFF
--- a/src/corona_hakab_model/main.py
+++ b/src/corona_hakab_model/main.py
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
     import pandas as pd
 
 logger = logging.getLogger('application')
-logger.setLevel("Info")
+logger.setLevel(logging.INFO)
 
 
 def main():
--- a/src/corona_hakab_model/main.py
+++ b/src/corona_hakab_model/main.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from argparse import ArgumentParser
 import matplotlib_set_backend
 import matplotlib.pyplot as plt
@@ -19,6 +20,9 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     import pandas as pd
 
+logger = logging.getLogger('application')
+logger.setLevel("Info")
+
 
 def main():
     parser = ArgumentParser("COVID-19 Simulation")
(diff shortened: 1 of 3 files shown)


src/corona_hakab_model/main.py
```diff
@@ -14,30 +16,58 @@
 from supervisor import LambdaValueSupervisable, Supervisable, SimulationProgression
 
 
-
 from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     import pandas as pd
 
+logger = logging.getLogger('application')
+logger.setLevel("Info")
```
RoyIsland, line 24: switch from "Info" to logging.INFO ("Info" gives me error when i run it)

src/corona_hakab_model/main.py
```diff
@@ -102,6 +132,56 @@ def main():
...
+    else:
+        circles_consts = CirclesConsts()
+
+    if args.matrix_consts_path:
+        matrix_consts = MatrixConsts.from_file(args.matrix_consts_path)
+    else:
+        matrix_consts = MatrixConsts()
+
+    gm = GenerationManger(circles_consts=circles_consts, matrix_consts=matrix_consts)
+    gm.save_to_folder(args.output_folder)
```
RoyIsland, line 157: save_to_folder not exists

RoyIsland:
`    if args.sub_command == "generate":
        with open(args.output, "wb") as w:
            write(gm.matrix_data.matrix, w)
        return`
this if will never be true

src/corona_hakab_model/main.py
```diff
@@ -102,6 +132,56 @@ def main():
...
+        generate_from_folder(args.consts_folder)
+        return
+    elif args.master_folder:
+        generate_from_master_folder(args.master_folder)
+        return
+
+    if not args.output_folder:
+        logger.error("No output folder given! use --output-folder")
+
+    if args.circles_consts_path:
```
RoyIsland, line 146: extract this part to function
`    if args.circles_consts_path:
        circles_consts = CirclesConsts.from_file(args.circles_consts_path)
    else:
        circles_consts = CirclesConsts()

    if args.matrix_consts_path:
        matrix_consts = MatrixConsts.from_file(args.matrix_consts_path)
    else:
        matrix_consts = MatrixConsts()`
Same code in 2 places

mor-m-shemesh: Done

src/corona_hakab_model/main.py
mor-m-shemesh: Done, I fixed it but overrode my own changes in a rebase

PR #194, opened 2020-04-12
Matrix analysis

### LINKED ISSUE #142: Add basic histograms
We should create A new module, fully separate from the whole simulation named MatrixAnalyzer.
It should be similar to the PopulationAnalyzer module described in: https://github.com/CoronaHakab/CoronaHakabModel/issues/140
It should contain:

היסטוגרמה של מספר קשרים כולל + ממוצע וחציון (מעגל ראשון כללי)

 - בהפרדה לסוגי קשרים
היסטוגרמה של מספר קשרי בית + ממוצע וחציון (מעגל ראשון בית)
היסטוגרמה של מספר קשרי עבודה + ממוצע וחציון
היסטוגרמה של מספר קשרי בית ספר + ממוצע וחציון
היסטוגרמה של מספר קשרי other + ממוצע וחציון

AmitGru:
This module plots histograms for connection types and exports the matrix raw data as csv file.
Closes #142


src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -21,14 +20,37 @@
...
 from generation.node import Node
+from bsa.parasym import write_parasym, read_parasym
+from bsa.scipy_sparse import read_scipy_sparse
 
 
 class MatrixData:
     __slots__ = ("matrix_type", "matrix", "depth")
 
+    # import/export variables
+    IMPORT_MATRIX_PATH = "../../output/matrix_data"
```
mor-m-shemesh, line 31: This type of path is dependent on where the script is ran from.
use project_structure.py to find the correct output folder path

bgmoshe: I agree with Mor, more over if you add new constant that is related to the project structure, please add it to project_structure.py

src/corona_hakab_model/main.py
```diff
@@ -88,6 +97,14 @@ def main():
     args, _ = parser.parse_known_args()
     set_seeds(args.seed)
 
+    if args.sub_command == 'analyze-matrix':
+        matrix_analyzer = MatrixAnalyzer(args.matrix_path)
+        matrix_analyzer.export_raw_matrices_to_csv()
+        matrix_analyzer.analyze_histograms()
+        if args.show:
```
bgmoshe, line 104: Why does analyze try to plot. Return a dataframe or series instead, and then let the user choose if he want to plot it or save it to a csv

src/corona_hakab_model/matrix_analysis/MatrixAnalysis.py
```diff
@@ -0,0 +1,135 @@
...
+        run over all types of connections and analyze the connection histogram
+        """
+        for connection_type in range(self.matrix_data.depth):
+            self.analyze_connection_histogram(connection_type)
+        self.analyze_connection_histogram()
+
+    def analyze_connection_histogram(self, connection_type=None):
+        if connection_type is None:
+            connection_matrix = sum(self.matrix_data.matrix)
+            connection_type = "all"
```
bgmoshe, line 68: Why not let this be the default value of the function instead?

src/corona_hakab_model/matrix_analysis/MatrixAnalysis.py
```diff
@@ -0,0 +1,135 @@
...
+            csv_writer = csv.writer(matrix_file, delimiter=',')
+            csv_writer.writerows([headers]+connections)
+
+    def analyze_histograms(self):
+        """
+        run over all types of connections and analyze the connection histogram
+        """
+        for connection_type in range(self.matrix_data.depth):
+            self.analyze_connection_histogram(connection_type)
+        self.analyze_connection_histogram()
```
bgmoshe, line 63: Please add comment that explains why you need this empty call

src/corona_hakab_model/matrix_analysis/MatrixAnalysis.py
```diff
@@ -0,0 +1,135 @@
...
+        max_connection = np.max(distribution_of_connections)
+        min_connection = np.min(distribution_of_connections)
+        connections_histogram, _ = np.histogram(distribution_of_connections,
+                                                bins=(max_connection-min_connection+1),
+                                                range=(min_connection - 0.5, max_connection + 0.5))
+        bins = np.arange(min_connection, max_connection+1)
+
+        self.export_histogram_to_csv(connections_histogram, bins,
+                                     average_connection, median_connection, connection_type)
+        self.plot_histogram(connections_histogram, bins, average_connection, median_connection, connection_type)
```
bgmoshe, line 86: Is seems like you bundle here things that shouldnt be bundled. Is sounds like this function just needs to return the statistics and then when calling export_hist.. it should receive histogram and save it.
Same goes for the plot

src/corona_hakab_model/generation/matrix_generator.py
AmitGru: תוקן

src/corona_hakab_model/matrix_analysis/MatrixAnalysis.py
AmitGru: תוקן

src/corona_hakab_model/matrix_analysis/MatrixAnalysis.py
AmitGru: תוקן

src/corona_hakab_model/matrix_analysis/MatrixAnalysis.py
AmitGru: קיבלתי את זה שצריך להפריד בין התפקידים של הפונקציות אבל לדעתי אין סיבה לתת ליוזר להחליט אם לייצא את הקבצים או לשמור את התמונות כי זאת כל המטרה של המודול... הוא מודול נפרד והוא אמור לרוץ רק כשרוצים לייצר את הדברים האלה, אז אין סיבה להעמיס על היוזר עם אופציות שונות להרצה

bgmoshe:
@AmitGru It would be super if you can talk with @mor-m-shemesh / @omri312 and see if this can become just a bunch of functions instead of class. We discussed this, I do not see any reason this should be a class

omri312:
> @AmitGru It would be super if you can talk with @mor-m-shemesh / @omri312 and see if this can become just a bunch of functions instead of class. We discussed this, I do not see any reason this should be a class

I think it's okay this will remain a class.
2 reasons - first is I think it's not that big of an issue right now, and in these places for now I preffer that we will run on with missions rather then block ourselves on relatively small decisions. This is of course a fine line, but for now I approve the way it is instead of starting changes.
Secondly - I personally do believe in classes over functions. There are many times that something that at the beginning looks like it will only be a bunch of functions, actually turns out to need to be a class by itself, and when that happens it is hard to change to that (while doing the opposite - using functions in classes - is pretty simple).

PR #199, opened 2020-04-12
Cleanup

benglewis:
Cleanup Tox messages


src/corona_hakab_model/sparse_matrix/sparse.py
```diff
@@ -5,15 +5,15 @@
 # the SWIG interface file instead.
```
bentheiii, line 5: This is an autogenerated file and tox shouldn't touch it

bgmoshe:
So we can close this PR ? @bentheiii @meirhalachmi

PR #200, opened 2020-04-12
Generation mw matrix

tal-nc:
generation now creates the new matrix.
p can be selected for work and school too


src/corona_hakab_model/analyzers/simulation_analysis.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    :param files: Files to compare. If not given, takes all the files in the output folder.
+    :return: N/A. For each parameter creates a file combining the information from each file.
+    """
+    if files is None:
+        files = glob(os.path.join(OUTPUT_FOLDER, "*.csv"))
+    dfs_dict = dict(map(lambda f: (os.path.basename(f),
+                                   pd.read_csv(f).drop(["Unnamed: 0"], axis=1, errors='ignore')),
+                        files))
```
bentheiii, line 19: use dictionary comprehension

src/corona_hakab_model/project_structure.py
```diff
@@ -0,0 +1,4 @@
+from os import path
```
bentheiii, line 1: this file gets a veto fro me, neither of these should be constant, what if the user wants to enter a different output folder? This makes some very dangerous assumptions about the file structure that will cause us misery down the line

src/corona_hakab_model/analyzers/simulation_analysis.py
tal-nc: these files aren't mine, they are from develop (had these open already before i switched to the new matrix branch and pulled from dev-4.0).

src/corona_hakab_model/project_structure.py
tal-nc: these files aren't mine, they are from develop (had these open already before i switched to the new matrix branch and pulling from dev-4.0).

tal-nc:
simulation analysis and project structure are pulled from develop. they aren't mine. i think they are fine, but if you want, we can remove them from this pr

bentheiii:
fine but i don't like em

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -52,36 +57,29 @@ def __init__(
         self.matrix_consts = matrix_consts
         self._unpack_population_data(population_data)
         self.size = len(self.agents)
-        self.depth = len(ConnectionTypes)
 
-        CoronaMatrix = corona_matrix.get_corona_matrix_class(matrix_consts.use_parasymbolic_matrix)
-        self.logger.info("Using CoronaMatrix of type {}".format(CoronaMatrix.__name__))
-        self.matrix = CoronaMatrix(self.size, self.depth)
+        # todo allow generic matrix type
```
bgmoshe, line 61: Is there reason this code is commented?

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -52,36 +57,29 @@ def __init__(
...
-                        con_type, self.social_circles_by_connection_type[con_type], current_depth
-                    )
-                elif con_type in Geographic_Clustered_types:
-                    self._create_community_clustered_circles_matrix(
-                        con_type, self.social_circles_by_connection_type[con_type], current_depth
-                    )
-                current_depth += 1
+        for con_type in ConnectionTypes:
+            if con_type in Connect_To_All_types:
+                self._create_fully_connected_layer(con_type, self.social_circles_by_connection_type[con_type])
```
bgmoshe, line 72: Why not save this functions in a dict (or named tuple) for each connection and then use comprehension instead of loops?

tal-nc: i am not sure i understand what you mean.

tal-nc:
this is broken in many places (update matrix and infection too).
the goal is to update them all to fit the new matrix one by one.

PR #206, opened 2020-04-13
Feature/moving parameters generator api

RoyIsland:


README.md
```diff
@@ -53,23 +53,26 @@
 - **Always run the simulator after making changes and before merging**, to make sure you didn't break anything.
 - Especially if not sure, **try to get at least 1 person's review before merging**.
 
-## Multiple Jsons generator
+## Multiple jsons generator
```
bgmoshe, line 56: Actually I thikn JSONs is more correct as JSON is not a word but abbriviation, but it doesn't really matter

RoyIsland: I agree, i'll change it back to Jsons

benglewis:
Do you want to create a JSON schema for these JSON files?

RoyIsland:
> Do you want to create a JSON schema for these JSON files?

For the moving parameters json file?

PR #207, opened 2020-04-13
multiply MW layer by factor

bentheiii:
allow MW to multiply by factor


src/corona_hakab_model/sparse_base.py
```diff
@@ -48,10 +48,12 @@ def __getitem__(self, item: Tuple[int, int]) -> Optional[Tuple[float, float]]:
         """
 
     @abstractmethod
-    def manifest(self, sample: np.ndarray = None) -> ManifestBase:
+    def manifest(self, sample: np.ndarray = None, global_prob_factor: float = 1) -> ManifestBase:
         """
         Create a manifested sparse matrix from sample rolls.
         if sample is None, a new rolls array is created
+        global_prob_factor denotes a global probability coefficient to apply to all probs. 0.5 means that each cell has
+        only half as much chance to manifest. A global_prob_factor of 0 is an error.
```
tal-nc, line 56: we do want to allow global prob factor of 0.
should check for it, and if so, switch sample to np.zeroes

src/corona_hakab_model/clustered_matrix/clustered.py
```diff
@@ -134,9 +134,11 @@ def __getitem__(self, item):
         l_j = self.local_indices[j]
         return self.clusters[c_i][l_i, l_j]
 
-    def manifest(self, sample=None):
+    def manifest(self, sample=None, global_prob_factor: float = 1):
         if sample is None:
             sample = generator.random(sum(c.size ** 2 for c in self.clusters), dtype=np.float32)
+        if global_prob_factor != 1:
```
tal-nc, line 140: check for 0 too, and if so switch sample to np.zeroes

src/corona_hakab_model/sparse_base.py
bentheiii: a. the sample should change to np.ones
b. I'd rather users check on their own because calling manifest with ones just adds overhead that could be avoided

bentheiii: I want to discourage calling this function with a factor zero because that can usually be optimized elsewhere

PR #213, opened 2020-04-13
Update readme and pipfile

idanmiara:


README.md
```diff
@@ -39,6 +39,15 @@
     - run: **python main.py all** to run both the generation and simulation.
     - run: **python main.py [simulate|generate] --help** for more help about each option.
     
+    
+## Installing and running under Linux (tested on Debian 10):
+- sudo apt install swig liblzma-dev libbz2-dev
+    - swig - building parasymoblic matrix needs swig
```
bentheiii, line 45: can't we pre-build the parasymbolic so that clients don't need swig to run it it?

README.md
```diff
@@ -39,6 +39,15 @@
...
+    - liblzma-dev libbz2-dev - since we use Pandas, we need Python with support for bz2 and lzma support
+        - (https://stackoverflow.com/questions/22346269/bz2-is-module-not-available-when-installing-pandas-with-pip-in-python-virtual) 
+- Building Python 3.8 (currently not on the official repos...)
+    - https://linuxize.com/post/how-to-install-python-3-8-on-debian-10/
```
bentheiii, line 49: can't we just make a shell script that installs all this?

README.md
idanmiara: Yes, I can push my build but I'm not sure it will work for every Linux...

README.md
idanmiara: I could!

README.md
bentheiii: you can push your CXX file and the clients won't need swig to build

idanmiara:
back to the drawing board

PR #217, opened 2020-04-13
add Dockerfile and linux scripts

idanmiara:
Docker file works but it gives an error about running qt in interactive mode.
Do we use any qt dependencies (i.e. matplotlib) that we need to run on the cloud ?
If not, let's remove that dependency so only the plotting module will be depend on it.


README.md
```diff
@@ -39,6 +39,15 @@
     - run: **python main.py all** to run both the generation and simulation.
     - run: **python main.py [simulate|generate] --help** for more help about each option.
     
+    
+## Installing and running under Linux (tested on Debian 10):
+- sudo apt install swig liblzma-dev libbz2-dev
+    - swig - building parasymoblic matrix needs swig
```
bentheiii, line 45: Using swig will require the users to also install swimport, which is a dev-only requirement. We cannot expect the client to install swig.

linux/deb_run.sh
```diff
@@ -0,0 +1,38 @@
...
+#https://linuxize.com/post/how-to-install-python-3-8-on-debian-10/
+sudo apt-get update -y && \
+sudo apt-get install -y build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libsqlite3-dev libreadline-dev libffi-dev curl liblzma-dev libbz2-dev && \
+curl -O https://www.python.org/ftp/python/3.8.2/Python-3.8.2.tar.xz && \
+tar -xf Python-3.8.2.tar.xz && \
+cd Python-3.8.2 && \
+./configure --enable-optimizations && \
+make -j `nproc` && \
+sudo make altinstall && \
+python3.8 --version
```
bentheiii, line 11: why not just call deb_install_python3.8.sh?

linux/deb_run.sh
```diff
@@ -0,0 +1,38 @@
...
+git clone https://github.com/CoronaHakab/CoronaHakabModel.git
+cd ./CoronaHakabModel/
+
+# we'll work with docker branch
+git checkout docker
+
+# install the project dependencies
+sudo pip3.8 install --upgrade pip
+sudo pip3.8 install pipenv
+pipenv install --dev
```
bentheiii, line 25: we can't expect client to install dev requirements

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    && apt-get install -y build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libsqlite3-dev libreadline-dev libffi-dev curl liblzma-dev libbz2-dev
+RUN curl -O https://www.python.org/ftp/python/3.8.2/Python-3.8.2.tar.xz \
+    && tar -xf Python-3.8.2.tar.xz
+RUN cd Python-3.8.2 \
+    && ./configure --enable-optimizations \
+    && make -j `nproc` \
+    && make altinstall \
+    && python3.8 --version \
+    && cd ..
+
```
benglewis, line 19: Why not just use the `python:3.8.2` Docker image?

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN apt-get install -y git
+RUN git clone https://github.com/CoronaHakab/CoronaHakabModel.git ${PROJDIR}
+
+WORKDIR ${PROJDIR}
+RUN git checkout docker
+
+# install the project dependencies
+RUN pip3.8 install --upgrade pip
+RUN pip3.8 install pipenv
+RUN pipenv install --dev
```
benglewis, line 33: Does it need the dev dependencies? Why?

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# RUN pipenv run python ${PROJDIR}/src/corona_hakab_model/parasymbolic_matrix/build_unix.py
+RUN cd ${PROJDIR}/src/corona_hakab_model/parasymbolic_matrix/ \
+    && pipenv run python build_unix.py
+
+#cd ~/CoronaHakabModel/
+#echo -e "export QT_QPA_PLATFORM=offscreen\nexport PYTHONPATH=`pwd`/src:`pwd`/src/corona_hakab_model\ncd ~/CoronaHakabModel/\npipenv shell" > ~/.bash_login
+#source ~/.bash_login
+
+# RUN apt-get install qt5-default
```
benglewis, line 49: Please delete commented code.

Pipfile
```diff
@@ -19,6 +19,7 @@ scikit-learn = "*"
 pyside2 = "*"
 matplotlib = "*"
 cached-property = "*"
+pandas = "*"
```
benglewis, line 22: Why is this in this PR?

Pipfile.lock
```diff
@@ -106,12 +106,34 @@
...
+                "sha256:6597df07ea361231e60c00692d8a8099b519ed741c04e65821e632bc9ccb924c",
+                "sha256:743bba36e99d4440403beb45a6f4f3a667c090c00394c176092b0b910666189b",
+                "sha256:858a0d890d957ae62338624e4aeaf1de436dba2c2c0772570a686eaca8b4fc85",
+                "sha256:863c3e4b7ae550749a0bb77fa22e601a36df9d2905afef34a6965bed092ba9e5",
+                "sha256:a210c91a02ec5ff05617a298ad6f137b9f6f5771bf31f2d6b6367d7f71486639",
+                "sha256:ca84a44cf727f211752e91eab2d1c6c1ab0f0540d5636a8382a3af428542826e",
+                "sha256:d234bcf669e8b4d6cbcd99e3ce7a8918414520aeb113e2a81aeb02d0a533d7f7"
+            ],
+            "index": "pypi",
+            "version": "==1.0.3"
```
benglewis, line 129: Likewise - why is this in this PR?

README.md
benglewis: Let’s release the compiled binaries to GitHub.
https://stackoverflow.com/questions/5207269/how-to-release-a-build-artifact-asset-on-github-with-a-script

docker.md
```diff
@@ -0,0 +1,23 @@
+# To build the image
```
benglewis, line 1: If you’re already going to do this, why not also add a quick link to a guide/set of instructions for how to install Docker.

linux/deb_run.sh
idanmiara: I was going to do that but then I was thinking that maybe this script would be ran outside the project so both scripts would need to be available side by side.
anyway I didn't test it yet as I'm working on making the Docker image work first

linux/deb_run.sh
idanmiara: OK.
but we need to think if if swimport is a dev package since we need for running if we don't deploy the binaries. I think building is the way to go instead of providing binaries as they might be different on different machines ? i.e. different arch or different cpu optimizations?

Dockerfile
idanmiara: I don't know what that image is based on, I feel comfortable using a Debian based distro.
Also I'm not sure that it includes the python dependencies that we need for Pandas.
Do you want to check it?

Dockerfile
idanmiara: As I replied to Ben, I'll check it out.

Dockerfile
idanmiara: ok

Pipfile
idanmiara: unrelated, but required as the project depends on it. didn't feel like making a PR just for that.

README.md
idanmiara: What's bad about making swimport and swig as regular dependencies instead of releasing binaries?

docker.md
idanmiara: It's a work in progress. didn't get to that yet.

idanmiara: I will. didn't get to it yet.

Pipfile.lock
idanmiara: same reason :)

Dockerfile
benglewis: It’s based upon Debian. You can install other dependencies - it just saves you the entire Python 3.8.2 compilation step and quite frankly it seems smarter to use the recommended Python Docker image which will be built to standards than to roll your own.

idanmiara:
OK I'll check it out.
I meant python build dependencies, if they are not present when python is
built then there is no support for them.

Pipfile
```diff
@@ -10,16 +10,16 @@ tox = "*"
...
 [packages]
 numpy = "*"
 scipy = "*"
 scikit-learn = "*"
+cached-property = "*"
 pyside2 = "*"
 matplotlib = "*"
-cached-property = "*"
 pandas = "*"
+swimport = "*"
```
bentheiii, line 22: I don't like swimport being necessary for running because it itself has other dependencies (lexicographic parsers for c++)

Pipfile
```diff
@@ -10,16 +10,16 @@ tox = "*"
 flake8 = "*"
 isort = "*"
 black = "*"
-swimport = "*"
 
 [packages]
 numpy = "*"
 scipy = "*"
 scikit-learn = "*"
+cached-property = "*"
```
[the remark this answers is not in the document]
idanmiara: true. removed.

Pipfile
idanmiara: from pipdeptree:
swimport==3.0.11
  - CppHeaderParser [required: >=2, installed: 2.7.4]
    - ply [required: Any, installed: 3.11]

I don't think that's too bad...

Pipfile
bentheiii: the project doesn't depend on it (at least it shouldn't), cached_property is standard lib

Pipfile
bentheiii: I still don't understand why we don't also add the .cxx file and then both swimport and swig would be unnecessary for building...

Dockerfile
idanmiara: done

Dockerfile
idanmiara: done

Dockerfile
idanmiara: done

PR #218, opened 2020-04-13
Add sparse matrix builds for Unix code and fix parasymbolic build again

benglewis:


benglewis:
I am merging this to allow continued development on macOS/Linux.
@idanmiara @tal-nc Please take a look and let me know if you think that any changes need to be made.

PR #233, opened 2020-04-14
Infection mw

tal-nc:
basic modifications for infections.py to fit the new matrix api.
note that still can't test this because magic operator is not yet defined.
small patch to sparse.py, untill build is updated


src/corona_hakab_model/infection.py
```diff
@@ -36,8 +36,9 @@ def _perform_infection(self) -> List[Agent]:
         # v = [True if an agent can infect other agents in this time step]
         v = np.random.random(len(self.manager.agents)) < self.manager.contagiousness_vector
 
+        # TODO save manifests in a queue-ish data structure
         # u = mat dot_product v (log of the probability that an agent will get infected)
-        u = self.manager.matrix.prob_any(v)
+        manifests, u = self.manager.matrix.POA(v)
```
bentheiii, line 41: rename manifests to `_` to show that it isn't being used

src/corona_hakab_model/sparse_matrix/sparse.py
```diff
@@ -23,6 +23,9 @@
 else:
     import _sparse
 
+# todo remove this and add generator to cpp build
+from consts import generator
```
bentheiii, line 27: add this line to build.py (even if you can't run it)

src/corona_hakab_model/infection.py
```diff
@@ -36,8 +36,9 @@ def _perform_infection(self) -> List[Agent]:
         # v = [True if an agent can infect other agents in this time step]
         v = np.random.random(len(self.manager.agents)) < self.manager.contagiousness_vector
```
bentheiii, line 37: v should be an array floats, maybe remove the random comparison?

src/corona_hakab_model/infection.py
tal-nc: done.

src/corona_hakab_model/sparse_matrix/sparse.py
tal-nc: done.
for now kept it both here and added it to build.py

src/corona_hakab_model/infection.py
tal-nc: good point. fixed.
maybe later contagiousness_vector should be renamed to V.

PR #245, opened 2020-04-15
detoxify

bentheiii:
removed all mentions of tox


idanmiara:
Once you go black you can't go back.
I think we should at least run it after we merge to dev. 😬

PR #249, opened 2020-04-15
Constrain initial agents

avivsimchony:
Added id's to circles and their exports, and added support for constraining the initial sick agents


src/corona_hakab_model/agent.py
```diff
@@ -80,18 +102,64 @@ def add_agent(self,agent_snapshot):
...
+            min_num,max_num = self.parse_range(row[connection_type.name])
+            social_circle_constraints.append(SocialCircleConstraint(min_num,max_num,connection_type))
+        return AgentConstraint(min_age,max_age,geographic_circle,social_circle_constraints)
+
+    def parse_range(self,range_element):
+        if type(range_element) is int or type(range_element) is float:
+            return range_element,range_element
+        assert type(range_element) is str
+        split_range = range_element.split(self.RANGE_DELIMITER)
+        assert 0< len(split_range) <= 2
```
bgmoshe, line 152: If you assert it is better that you print something meaningfull

src/corona_hakab_model/agent.py
```diff
@@ -80,18 +102,64 @@ def add_agent(self,agent_snapshot):
...
+        if len(split_range) == 1:
+            return self.parse_str_to_num(split_range[0]),self.parse_str_to_num(split_range[0])
+        return self.parse_str_to_num(split_range[0]), self.parse_str_to_num(split_range[1])
+
+    @staticmethod
+    def parse_str_to_num(val):
```
bgmoshe, line 158: 2 things:
1) Why this is part of this class?
2) Why not just try to convert to float.

src/corona_hakab_model/generation/circles.py
```diff
@@ -1,11 +1,13 @@
...
+import pandas as pd
 
 class Circle:
-    __slots__ = "kind", "agent_count"
+    __slots__ = "kind", "agent_count", "circle_id"
 
-    def __init__(self):
+    def __init__(self,circle_id=-1):
         self.agent_count = 0
+        self.circle_id = circle_id
```
bgmoshe, line 10: Why is this optional?

src/corona_hakab_model/generation/circles.py
```diff
@@ -55,10 +56,33 @@ def get_indexes_of_my_circle(self, my_index):
...
+    id: int
+
+class SocialCircleConstraint:
+    def __init__(self,min_members,max_members,connection_type):
+        self.min_members = min_members
+        self.max_members = max_members
+        self.connection_type = connection_type
+
+    def meets_constraint(self,agent):
+        assert agent is not None
```
bgmoshe, line 75: Why? Let the normal error jump if agent is None...

src/corona_hakab_model/generation/circles.py
```diff
@@ -55,10 +56,33 @@ def get_indexes_of_my_circle(self, my_index):
...
+        has_circle = False
+        if pd.isna(self.min_members) and pd.isna(self.max_members):
+            return  True
+        try:
```
bgmoshe, line 79: Why not use boolean variables?

avivsimchony:
> Some work is needed
Very true, fixing issues and creating new pull request

src/corona_hakab_model/agent.py
```diff
@@ -69,9 +70,30 @@ class AgentSnapshot:
     social_circles: list
 
 
+class AgentConstraint:
+    MAX_AGE = 120
+    def __init__(self,min_age=0,max_age=MAX_AGE,geographic_circle="",social_circle_constraints=None):
```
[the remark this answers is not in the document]
avivsimchony: Mistake, fixing

src/corona_hakab_model/agent.py
```diff
@@ -69,9 +70,30 @@ class AgentSnapshot:
     social_circles: list
 
 
+class AgentConstraint:
+    MAX_AGE = 120
```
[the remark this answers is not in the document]
avivsimchony: Fun Fact! Also this parameter is unnecessary, so removed

src/corona_hakab_model/agent.py
```diff
@@ -69,9 +70,30 @@ class AgentSnapshot:
...
+    def __init__(self,min_age=0,max_age=MAX_AGE,geographic_circle="",social_circle_constraints=None):
+        self.min_age = min_age
+        self.max_age = max_age
+        self.geographic_circle = geographic_circle
+        self.social_circle_constraints = social_circle_constraints
+
+    def meets_constraint(self,agent: AgentSnapshot):
+        try:
```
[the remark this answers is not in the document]
avivsimchony: I thought it was rather elegant, but I see your point, fixing

src/corona_hakab_model/agent.py
```diff
@@ -69,9 +70,30 @@ class AgentSnapshot:
...
+            assert pd.isna(self.min_age) or agent.age >= self.min_age
+            assert pd.isna(self.max_age) or agent.age <= self.max_age
+            assert self.geographic_circle is None or self.geographic_circle == agent.geographic_circle
```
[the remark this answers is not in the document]
avivsimchony: should be nan, fixing (also removed "" init, that was a mistake

src/corona_hakab_model/agent.py
```diff
@@ -69,9 +70,30 @@ class AgentSnapshot:
...
+            if self.social_circle_constraints is not None:
+                for constraint in self.social_circle_constraints:
+                    assert constraint.meets_constraint(agent)
```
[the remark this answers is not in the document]
avivsimchony: I'm not sure where exactly you mean, but this entire are was refactored, so should be fine

src/corona_hakab_model/agent.py
```diff
@@ -80,18 +102,64 @@ def add_agent(self,agent_snapshot):
...
+    RANGE_DELIMITER = '~'
+    def __init__(self,constraints_file_path=None):
+        self.constraints = self.parse_constraints(constraints_file_path)
+
+    def parse_constraints(self,constraints_file_path):
+        if constraints_file_path is None:
+            return None
+        df_constraints = pd.read_csv(constraints_file_path)
+        constraints = []
+        for index, row in df_constraints.iterrows():
```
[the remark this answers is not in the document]
avivsimchony: MUCH prettier, don't know how I missed that

src/corona_hakab_model/agent.py
```diff
@@ -80,18 +102,64 @@ def add_agent(self,agent_snapshot):
...
+        min_age,max_age = self.parse_range(row[self.AGE])
+        geographic_circle = row[self.GEOGRAPHIC_CIRCLE]
+        social_circle_constraints = []
+        for connection_type in ConnectionTypes:
+            min_num,max_num = self.parse_range(row[connection_type.name])
+            social_circle_constraints.append(SocialCircleConstraint(min_num,max_num,connection_type))
+        return AgentConstraint(min_age,max_age,geographic_circle,social_circle_constraints)
+
+    def parse_range(self,range_element):
+        if type(range_element) is int or type(range_element) is float:
```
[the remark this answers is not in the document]
avivsimchony: The idea was to treat numbers and strings differently, due to both options being possible with pandas, and both ints and floats could come out. execution was sloppy though, changed to using isinstance, and only testing for string.

src/corona_hakab_model/agent.py
avivsimchony: Fixed to exception with meaningful message

src/corona_hakab_model/agent.py
avivsimchony: A lot of these parameters are integers, so I do prefer to keep them that way if possible, just in case. Relocating it sounds sensible, moved to util.

src/corona_hakab_model/generation/circles.py
avivsimchony: currently not every circle needs an id, and assigning a unique id to every circle is a pretty daunting task. This might be the way to go in the future, but it doesn't really solve any current problems, since the id's are really only useful for exporting to humans, and might create new ones.

src/corona_hakab_model/generation/circles.py
avivsimchony: Done.

src/corona_hakab_model/generation/circles.py
avivsimchony: Fixed.

src/corona_hakab_model/agent.py
```diff
@@ -80,18 +102,64 @@ def add_agent(self,agent_snapshot):
...
+            return range_element,range_element
+        assert type(range_element) is str
```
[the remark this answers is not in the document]
avivsimchony: no longer asserting there

bgmoshe:
@avivsimchony  Please resolve the conflicts

PR #251, opened 2020-04-16
Agents df

mishana:


src/corona_hakab_model/agents_df.py
```diff
@@ -0,0 +1,114 @@
...
+    def susceptible(self):
+        return np.array([ms.susceptible for ms in self.medical_state])
+
+    @property
+    def detectable(self):
+        return np.array([ms.detectable for ms in self.medical_state])
+
+    @property
+    def test_willingness(self):
+        return np.array([ms.test_willingness for ms in self.medical_state])
```
michaeleh, line 99: for test willingness, detectable, susceptible and contagiousness it would be better to save a vector for each and not create a list every time we call that function, would make a difference for large number of agents

src/corona_hakab_model/manager.py
```diff
@@ -121,32 +111,35 @@ def step(self):
...
-                    self.new_detected_daily += 1
 
-            agent.set_test_result(test_result)
+        if new_results:
+            agent_indices, test_results, _ = zip(*new_results)
+            agent_indices = list(agent_indices)  # can't access numpy array with a tuple of indices
+            self.agents_df.set_tests_results(agent_indices, test_results)
+            tests_taken_mask = np.array([res != DetectionResult.NOT_TAKEN for res in test_results])
+            self.new_detected_daily = np.count_nonzero(
+                tests_taken_mask & self.agents_df.ever_tested_positive[agent_indices])
```
michaeleh, line 121: ever tested positive is ever detected not new detected

michaeleh, line 121: ever tested positive might not be new detection

mishana: fixed, if our definition of new detected is ok

src/corona_hakab_model/agent.py
```diff
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
 from dataclasses import dataclass
-from generation.connection_types import ConnectionTypes
+from typing import TYPE_CHECKING
```
bgmoshe, line 4: Preferably do not do this,  differences when they are not needed make conflicts and unneded diffs

PR #254, opened 2020-04-16
Added support of exporting sick agents to csv periodically

erangold:
Defined a new type of supervisable (TabularSupervisable), for exporting tabular samples to csv files (e.g. infected agents, in this particular case).


src/corona_hakab_model/supervisor.py
```diff
@@ -217,6 +259,28 @@ def name(self) -> str:
         return self.__name
 
 
+class _CurrentInfectedTable(TabularSupervisable):
+    def __init__(self, interval):
+        super().__init__(interval)
+        self.sick_states_names = ['Latent', 'Asymptomatic', 'Symptomatic', 'Hospitalized', 'ICU']
```
bgmoshe: Do not use this, the states can change (there is an active PR for that)
When they change you also need to update it here...

src/corona_hakab_model/supervisor.py
```diff
@@ -217,6 +259,28 @@ def name(self) -> str:
...
+
+    def get(self, manager) -> Dict[str, List]:
+        agent_ids = []
+        medical_status = []
+
+        for state_name in self.sick_states_names:
```
bgmoshe: Manager has medical state machine from which you can access all the states

erangold: I need access to the states which indicate any type of sickness. Can I assume that all medical states, except the initial state, are such?

bgmoshe: No, Deceased and Recovered are not such states

PR #255, opened 2020-04-16
Feature/update medical state machine

mor-m-shemesh:
Updated the medical state machine according to the new model:
https://docs.google.com/spreadsheets/d/1el3BT8fX6g7d65QbNpxnalGsKmnW5cy9Ob9B_1Bch34/edit?usp=sharing

This PR has a few caveats - 

1. To handle the Latent-presymptomatic coupling, I split Latent into two branching states.
The symptomatic latent state leads to the presymtomatic state and thus is shorter in duration.

2. the markovian function doesn't like true discrete distribution, so I estimated the actual data into the dist() util function.

3. All of the age-dependant states were averaged according to the actual population's age distribution. In 4.0 the data should instead use the AgentAwareState to account for difference in age.

ALSO - 
I spent a LOT of time formatting and fixing the data, so I left it in comments below the approximated fields.


bgmoshe:
image

src/corona_hakab_model/consts.py
```diff
@@ -37,22 +37,45 @@ class Consts(NamedTuple):
...
+    asymptomatic_to_pre_recovered_days: rv_discrete = dist(10, 18, 35)
+    # Actual distribution: rv_discrete(values=(
+    # [10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35],
+    # [0.013,0.016,0.025,0.035,0.045,0.053,0.061,0.065,0.069,0.069,0.065,0.063,0.058,0.053,0.056,0.041,0.040,0.033,
+    # 0.030,0.025,0.020,0.015,0.015,0.015,0.010,0.010]))
     # infections ratios
+    pre_symptomatic_infection_ratio: float = 0.75
+    mild_condition_infection_ratio: float = 0.40
     symptomatic_infection_ratio: float = 0.75
     asymptomatic_infection_ratio: float = 0.25
```
meirhalachmi, line 80: The real asymptomatic infection ratio (according to the latest model) is 0.
Wanna change it instead of opening a new PR?

src/corona_hakab_model/consts.py
```diff
@@ -273,27 +318,100 @@ class ImmuneTerminalState(ImmuneState, TerminalState):
...
+            probability=...
+        )
+
+        pre_recovered.add_transfer(
+            recovered,
+            duration=self.pre_recovered_to_recovered_days,
+            probability=...
+        )
+
+        asymptomatic.add_transfer(
```
meirhalachmi, line 410: Asymp should go directly to recovered, not to pre-recovered.

src/corona_hakab_model/consts.py
```diff
@@ -245,26 +254,62 @@ class ContagiousStochasticState(ContagiousState, StochasticState):
...
+        # latent-asymp - will have the durations of latent, summed by each day
+        # probability for each is same as probability from latent to presymp and asymp
+
         susceptible = SusceptibleTerminalState("Susceptible", test_willingness=self.susceptible_test_willingness)
         latent = ImmuneStochasticState("Latent", detectable=False, test_willingness=self.latent_test_willingness)
-        silent = ContagiousStochasticState(
-            "Silent", contagiousness=self.silent_infection_ratio, test_willingness=self.silent_test_willingness
+        latent_presymp = ImmuneStochasticState(
+            "Latent-Presymp",
+            detectable=False,
```
meirhalachmi, line 267: That's one of the problems with this simplification - you should be detectable in presymp, and undetectable in latent, and here you had to choose only one detectability option for this mesh-up state.

Maybe we should ask someone whether we should make this state detectable?

mor-m-shemesh: Good point, I am talking to Amit Spinrad today so I'll ask him

src/corona_hakab_model/consts.py
mor-m-shemesh: Fixed

src/corona_hakab_model/consts.py
mor-m-shemesh: Fixed

src/corona_hakab_model/consts.py
meirhalachmi: I'm resolving this comment. If needed, the constant will be changed as a hotfix.

PR #256, opened 2020-04-16
Issue #162 JSON load Consts

### LINKED ISSUE #162: JSON to state machine
**Point of contact**: Mor Shemesh

We currently set our medical state machine in consts.py.
We would like to be able to dynamically and easily configure the state machine parameters (and perhaps in second step the whole state machine) from a JSON file.

Create and define a new JSON file with the relevant parameters, and enable importing it and creating a state machine from it.

benglewis:
Note: This is already in dev-4.0

Partially resolves #162
Correct typos, make const classes dataclasses rather than NamedTuples so that they can be initialized with other values.
Migrate code from using from_file method to using from_json and modify and re-create the Python file as a new JSON file for loading the consts.


bgmoshe:
@meirhalachmi  if @benglewis doesn't work on it anymore should we close this issue or reassign it?

PR #262, opened 2020-04-16
fix weekly/daily connections

rlfreiman:
Two connected agents could have the connection as daily for one, and weekly for the other (that is 1/7 strength).
This PR fixes it to be symmetric at the temporary cost of some memory

### DIFF
--- a/src/corona_hakab_model/generation/matrix_generator.py
+++ b/src/corona_hakab_model/generation/matrix_generator.py
@@ -228,6 +228,8 @@ class MatrixGenerator:
         self._add_small_circle_connections(super_small_circles_combined, connections, total_connections_float)
 
         # insert all connections to matrix
+        # we need to remember the strengths so the connection will be symmetric
+        known_strengths = {}
         for agent, conns in zip(self.agents, connections):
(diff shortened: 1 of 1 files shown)


src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -237,6 +239,15 @@ def _create_random_clustered_circles_matrix(self, con_type: ConnectionTypes, cir
             strengthes = np.random.choice(
                 [connection_strength, connection_strength / 7], size=len(conns), p=[daily_share, weekly_share]
             )
+            # check if some strengths were determined earlier
+            for conn in conns:
+                known_strength = known_strengths.get((conn, agent.index), None)
+                if known_strength != None:
+                    strengthes[np.where(conns==conn)] = known_strength
```
tal-nc, line 246: wouldn't it be easier to enumerate conns and just address strengths by the index?
for i, conn in conns:
    strengths[i] = ...

anyway if this np.where works, i guess it doesnt really matters.

tal-nc (approved):

PR #264, opened 2020-04-17
Sanity check zero time

### LINKED ISSUE #247: Add sanity checks - can't have 0 time for state
We want to not allow inserting 0 as a time for a state,
Or insert a distribution of times where 0 is possible.

Make a check that return an error to the user if he accidently inserted that.

denzo9:
Added a check for the possibility of zero values in time distributions provided by consts.


src/corona_hakab_model/main.py
```diff
@@ -79,6 +79,17 @@ def run_simulation(args):
         consts = Consts.from_file(args.simulation_parameters_path)
     else:
         consts = Consts()
+
+    # Sanity check for medical states time
+    time_parameters = ["latent_to_pre_symptomatic_days", "latent_to_asymptomatic_days",
```
ghost: maybe extract to other file

src/corona_hakab_model/main.py
```diff
@@ -79,6 +79,17 @@ def run_simulation(args):
...
+                       "pre_symptomatic_to_mild_condition_days", "mild_to_close_medical_care_days",
+                       "mild_to_need_icu_days", "mild_to_pre_recovered_days", "close_medical_care_to_icu_days",
+                       "close_medical_care_to_mild_days", "need_icu_to_deceased_days", "need_icu_to_improving_days"]
+
+    for time_param in time_parameters:
+        if getattr(consts, time_param).pmf(0) != 0:
```
bentheiii: use util.lower_bound

src/corona_hakab_model/main.py
bentheiii: I don't like these hardcoded names. Maybe iterate over all the stochastic states in the machine?

bentheiii (changes requested):

src/corona_hakab_model/main.py
denzo9: Maybe perform the check when a transfer is added to a stochastic state?

denzo9: Moved the check inside state_machine. This eliminates the ugly hardcoded names, but the check is performed later. In my opinion definitely with it

denzo9: Moved the check inside state_machine

src/corona_hakab_model/main.py
denzo9: fixed, switched to lower_bound

bentheiii (approved):

PR #265, opened 2020-04-17
Feature/output directory fix

ghost:


src/corona_hakab_model/project_structure.py
```diff
@@ -1,4 +1,5 @@
 from os import path
 
-MODEL_FOLDER =  path.join(path.dirname(path.abspath(__file__)), "..", "..")
+MODEL_FOLDER = path.join(path.dirname(path.abspath(__file__)), "..", "..")
 OUTPUT_FOLDER = path.join(MODEL_FOLDER, "output")
+SIM_OUTPUT_FOLDER = path.join(OUTPUT_FOLDER, "sim_records")
```
NoamShapira, line 5: I Think, we should add documentation of the difference between the 3 folders

NoamShapira (approved):

PR #268, opened 2020-04-17
added ability to validate matrix

### LINKED ISSUE #241: Add sanity check - all values in w are probabilities
Note that this is only for 3.5.

All values in W **After Normalization** should be  negative, as they are ln(1-p), where p is a probability.
Also, all values in the containgusness vector all values should be probabilities, i.e. between 0 and 1.

Add checks in the code (in matrix creation, in updating, and/or in the matrix itself), that makes sure these values aren't accidently something else.

If during the checks we find that it does happen - please open a bug with the details and contact me (Omer).

rlfreiman:
It is pretty fast, so i added it by default.
Checked it - found bug (fixed elsewhere) and finds induced bugs (asymmetric connections, or validating before normalizing).
closes #241


bentheiii (changes requested):
thous should be a python method that instead returning true/false, should raise an exception. Putting it in c+ doesn't help with performance and only obfuscate code.

rlfreiman:
In a population of 20000 there is a few second difference in validation between doing it in python and c++ (unless i'm doing it wrong)
I uploaded another PR, in python. once one is merged the other can be deleted.

rlfreiman:
Closed - alternative PR #269 was merged.

PR #269, opened 2020-04-17
validate parasymbolic matrix in python

### LINKED ISSUE #241: Add sanity check - all values in w are probabilities

rlfreiman:
validate parasymbolic matrix in python via existing c++ code

### DIFF
--- a/src/corona_hakab_model/update_matrix.py
+++ b/src/corona_hakab_model/update_matrix.py
@@ -48,7 +48,7 @@ class UpdateMatrixManager:
         self.total_contagious_probability = None
         self.normalize()
         self.validate_matrix()
-
+        
     def normalize(self):
         """
         this function should normalize the weights within W to represent the infection rate.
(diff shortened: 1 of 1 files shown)


src/corona_hakab_model/update_matrix.py
```diff
@@ -125,3 +126,11 @@ def apply_full_isolation_on_agent(self, agent):
         for connection_type in ConnectionTypes:
             self.matrix.mul_sub_row(connection_type, agent.index, factor)
             self.matrix.mul_sub_col(connection_type, agent.index, factor)
+
+    def validate_matrix(self):
+        submatrixes_rows_nonzero_columns = self.matrix.non_zero_columns()
+        for rows_nonzero_columns in submatrixes_rows_nonzero_columns:
+            for row_index, nonzero_columns in enumerate(rows_nonzero_columns):
+                for column_index in nonzero_columns:
+                    assert self.matrix.get(row_index, column_index) == self.matrix.get(column_index, row_index)
```
bgmoshe: Please instead of asserting return the value of the validation.
This can be usefull if we will want to take some measures =/= asserting on spot.
The code that uses validate_matrix can assert

bgmoshe: Also, when asserting please give the exact reason for which the matrix is not valid.

bentheiii (approved):

tal-nc (approved):

bgmoshe (changes requested):
Small fix

bgmoshe (approved):

PR #271, opened 2020-04-17
Feature/parameters per geographic circle

### LINKED ISSUE #266: Split generating and parameters per geographic-circle
Currently we're generating all agents based on global parameters (though they are each generated by their geographic-circle).
We want it to be possible to give all of the relvant parameters **Per Geographic circle** - for example, I'll have some age distribution for the north, and another for the south.

ghost:


src/corona_hakab_model/generation/circles_consts.py
```diff
@@ -74,33 +109,32 @@ def from_file(cls, param_path):
...
+                geo_circle["teachers_workforce_ratio"]
             )
-            for i in range(self.geo_circles_amount)
+            for i, geo_circle in enumerate(self.geo_circles)
         ]
 
-    def get_age_distribution(self):
-        return rv_discrete(10, 70, values=(self.ages, self.age_prob))
+    def get_age_distribution(self, geo_circle):
+        return rv_discrete(10, 70, values=(geo_circle["ages"], geo_circle["age_prob"]))
```
tal-nc: 10, 70 could be switched with: min /max (geo_circle["ages"]).

tal-nc (approved):
looks great. i think it makes the code a lot more read-able.

src/corona_hakab_model/generation/circles_consts.py
bgmoshe: @tal-nc  Why have those values at all if you specifies values?
@RoyIsland

tal-nc: good point. it isn't necessary.
tbh we don't even use this range anywhere in the code (as far as i know).

PR #275, opened 2020-04-18
Added kindergartens

### LINKED ISSUE #273: Add kindergardens
We would like to support kindergardens.
This should be a new type of connection.
It will be very similiar to schools (only sizes entered will probably be much smaller).

**Parameters to start with**
- 10% of children (age "10" for now) should be in kindergardens.
- Sizes should be either 10 or 20
- 0.002 of workforce should be kindergardneres.

**Important notes**
- Children should go _either to school or to kindergarden, not both._
- Adults should be working _either in school, in kindergarden or in a work_
- In the matrix generation, kindergardens should behave exactly like schools.
- Add to consts.py policies for kindergardens (closing and opening) as there are for works and schools, and make sure they work as expected.

denzo9:


src/corona_hakab_model/generation/circles_consts.py
```diff
@@ -54,21 +59,25 @@ class CirclesConsts(NamedTuple):
             "ages": [10, 40, 70],
             "age_prob": [0.30, 0.45, 0.25],
             "teachers_workforce_ratio": 0.04,  # ratio of teachers out of workforce
+            "kindergarten_workforce_ratio": 0.002,  # ratio of kindergarten workers out of workforce
```
rlfreiman: Did you check it's enough to populate the kindergardens?

src/corona_hakab_model/generation/circles_consts.py
```diff
@@ -54,21 +59,25 @@ class CirclesConsts(NamedTuple):
...
             "connection_type_prob_by_age_index": [
                 {
                     ConnectionTypes.Work: 0,
-                    ConnectionTypes.School: 0.95,
+                    ConnectionTypes.Kindergarten: 0.1,
```
rlfreiman, line 66: kindergarDen

src/corona_hakab_model/generation/circles_consts.py
```diff
@@ -22,21 +22,25 @@ class CirclesConsts(NamedTuple):
...
+                    ConnectionTypes.School: 0.85,
                     ConnectionTypes.Family: 1.0,
                     ConnectionTypes.Other: 1.0,
                 },
                 {
                     ConnectionTypes.Work: 0.9,
+                    ConnectionTypes.Kindergarten: 0,
```
bgmoshe, line 36: What about kindergarten teachers?

denzo9: If I understand the question correctly, it is the same as with schools. The connection type for kindergarten workers is ConnectionTypes.Kindergarten

src/corona_hakab_model/generation/geographic_circle.py
```diff
@@ -34,22 +36,34 @@ def generate_agents_ages_and_connections_types(self):
...
-                if np.random.random() < self.data_holder.teachers_workforce_ratio:
-                    agent_connection_types.remove(ConnectionTypes.Work)
-                    agent_connection_types.append(ConnectionTypes.School)
+            if agent.is_adult() and np.random.random() < self.data_holder.connection_types_prob_by_age[age][
+                ConnectionTypes.Work]:
+
+                # if the agent works, decide workplace
+                agent_connection_types.append(workplace_distribution.rvs())
+
+            education_prob = [self.data_holder.connection_types_prob_by_age[age][connection_type] for
```
rlfreiman: this only needs to be calculated for a non-adult, move it inside the IF below

src/corona_hakab_model/generation/geographic_circle.py
```diff
@@ -34,22 +36,34 @@ def generate_agents_ages_and_connections_types(self):
...
+                              connection_type in Education_Types]
+            education_type = rv_discrete(values=([Education_Types + [-1], education_prob + [1 - sum(education_prob)]])).rvs()
```
rlfreiman: instead of the "+ [-1]", why not compare random() with sum(education_prob) in the IF below? much more readable

rlfreiman (changes requested):

src/corona_hakab_model/generation/geographic_circle.py
```diff
@@ -34,22 +36,34 @@ def generate_agents_ages_and_connections_types(self):
...
+        workplace_distribution = rv_discrete(
+            values=([ConnectionTypes.School, ConnectionTypes.Kindergarten, ConnectionTypes.Work],
+                    [self.data_holder.teachers_workforce_ratio, self.data_holder.kindergarten_workforce_ratio,
+                    1 - self.data_holder.teachers_workforce_ratio - self.data_holder.kindergarten_workforce_ratio])
+        )
         for agent, age in zip(self.agents, ages):
             agent.age = age
             agent_connection_types = []
-            for connection_type in ConnectionTypes:
+            for connection_type in Non_Exclusive_Types:
```
rlfreiman: I an not sure that Non_Exclusive_Types is very clear. rather it's better to explicitly handle Work/School/Kindergarden, and all connection types not in [Work/School/Kindergarden].

src/corona_hakab_model/generation/circles_consts.py
denzo9: That's the parameter given in the issue, so I didn't specifically check that. Will do

src/corona_hakab_model/generation/geographic_circle.py
denzo9: That was my original implementation. This will require normalizing the probabilities, but I guess it is more readable

src/corona_hakab_model/generation/geographic_circle.py
denzo9: In my opinion, the NonExclusive_type is slightly more readable

rlfreiman (approved):

PR #277, opened 2020-04-18
Feature/270 daily infection counter by type

### LINKED ISSUE #270: For each step - how many people were infected of each connection type
For each new person that got infected -
- Take the probability that he got infected **from each one of the submatrixes** (i.e work, home, family).
- "Roll a weighted die" to decide which connection he practically got infected through.

Example:
Say an agent got infected. I calculate the probability he'll get infected from work and get 0.1, from his family I get 0.3, and from other 0.15.
I now "roll a die" with chances that have the ratio 0.1 : 0.3 : 0.13 to get if he was infected from his work : family : other.

Then - add this to a counter of how many people got infected today from each of the connection types.
Export the data inside our main output csv.

rlfreiman:
closes #270. depends on PR 275!

### DIFF
--- a/src/corona_hakab_model/infection.py
+++ b/src/corona_hakab_model/infection.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Dict, List
 
 import numpy as np
 from agent import Agent
+from generation.connection_types import ConnectionTypes
 
 if TYPE_CHECKING:
     from medical_state import MedicalState
@@ -43,7 +44,13 @@ class InfectionManager:
 
         infections = self.manager.susceptible_vector & (np.random.random(u.shape) < u)
         infected_indices = np.flatnonzero(infections)
-
+        
+        # get infection methods
+        infection_methods = []
+        nonzero_columns = self.manager.matrix.non_zero_columns()
+        for agent_index in infected_indices:
+            infection_methods.append(self._get_infection_method(agent_index, nonzero_columns))
+            
         # new_infected: dict -
         # key = medical state (currently only susceptible state which an agent can be infected)
         # value = list of agents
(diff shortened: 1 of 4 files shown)


src/corona_hakab_model/infection.py
```diff
@@ -52,4 +59,18 @@ def _perform_infection(self) -> List[Agent]:
...
             new_infected.append(agent)
 
-        return new_infected
+        return new_infected, infection_methods
+
+    def _get_infection_method(self, agent_index, non_zero_columns):
+        probability_by_conn_type = []
+        for connection_type in ConnectionTypes:
+            possible_infectors = non_zero_columns[connection_type][agent_index]
+            conn_infection_probability = sum([self.manager.matrix.get(
```
tal-nc: i don't think that summing works here.
i think it should be:
..._probability = 1 - prod ([1 - self.manager.matrix.get(...) for ...])

the possibility to get infected from 3 meetings isn't the sum of the possibility to get infected in each meeting, but is 1 - the possibility to not get infected at all.

src/corona_hakab_model/infection.py
```diff
@@ -52,4 +59,18 @@ def _perform_infection(self) -> List[Agent]:
             agent = self.manager.agents[index]
             new_infected.append(agent)
 
-        return new_infected
+        return new_infected, infection_methods
+
+    def _get_infection_method(self, agent_index, non_zero_columns):
```
tal-nc: this method ignores the medical states of the possible infectors.
should use the contagiousness vector either in selecting possible infectors, or in the probability calculation

tal-nc (changes requested):

src/corona_hakab_model/infection.py
rlfreiman: fixed, now each infection probability is separate

src/corona_hakab_model/infection.py
rlfreiman: fixed

PR #278, opened 2020-04-19
updated implementation to use pathlib instead of os.path

erangold:


src/corona_hakab_model/supervisor.py
```diff
@@ -274,7 +271,7 @@ def get(self, manager: "manager.SimulationManager"):
         pass
 
     def snapshot(self, manager: "manager.SimulationManager"):
-        if manager.current_step % self.interval == 0:
+        if manager.current_step > 0 and manager.current_step % self.interval == 0:
```
bgmoshe: Why skip 0? It is the day of the infection. If we want to snapshot it you are disabling this ability

src/corona_hakab_model/supervisor.py
```diff
@@ -47,13 +47,10 @@ def snapshot(self, manager):
             s.snapshot(manager)
 
     def dump(self, filename=None):
-        file_name = Path(filename) if filename else Path(SIM_OUTPUT_FOLDER) / (datetime.now().strftime("%Y%m%d-%H%M%S") + ".csv")
+        output_folder = Path(SIM_OUTPUT_FOLDER)
```
bgmoshe: Why do you use Path here? SIM_OUTPUT_FOLDER is supposed to be a path not str

bgmoshe (commented):
Small changes

src/corona_hakab_model/supervisor.py
ghost: SIM_OUTPUT_FOLDER is already a path object, so you can use it as is

src/corona_hakab_model/supervisor.py
erangold: I thought that 0 won't be so informative, so I skipped it.. I added again in the latest commit

ghost (approved):

src/corona_hakab_model/supervisor.py
erangold: Okay, changed it in the latest commit

PR #281, opened 2020-04-19
added permutation to the first sick agents by default

### LINKED ISSUE #280: Randomize the initial sick agents by default
Make the first sick agents chosen (whether constrained or unconstrained) random instead of the first in the list, which makes them overly connected.

avivsimchony:
for issue 280, see slack conversations

### DIFF
--- a/src/corona_hakab_model/manager.py
+++ b/src/corona_hakab_model/manager.py
@@ -1,7 +1,7 @@
 import logging
 from collections import defaultdict
 
-from random import random, choice
+from random import shuffle
 from typing import Callable, Dict, Iterable, List, Union
 import infection
 import update_matrix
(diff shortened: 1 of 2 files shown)


bgmoshe (approved):

src/corona_hakab_model/manager.py
```diff
@@ -159,13 +159,21 @@ def setup_sick(self):
         """
         agents_to_infect = []
         agent_index = 0
+        agent_permutation = list(range(len(self.agents)))
+
+        if self.run_args.randomize:
+            self.logger.info("creating permutation")
+            shuffle(agent_permutation) #this is somewhat expensive for large sets, but imho it's worth it.
```
bgmoshe, line 166: For large sets you can split them to smaller sets and shuffle each one separately. For this you should look at what size is performance an issue. Lets say we have 100,000 agents, then if we shuffle them in buckets of 10,000 I think we should be okay.

avivsimchony: We could, but shuffling 10,000,000 agents took about 10 seconds on my laptop, so I think that would be a premature optimization.

PR #282, opened 2020-04-19
fix all paths relative to project_structure.py paths!

idanmiara:


src/corona_hakab_model/analyzers/config.py
```diff
@@ -1,6 +1,4 @@
 from project_structure import OUTPUT_FOLDER
-from pathlib import Path
 
-TIME_OUTPUT_PATH = Path(__file__).parent / "../../../output"
-SIM_OUTPUT_FOLDER = Path(__file__).parent / "../../../output/sim_records"
+TIME_OUTPUT_PATH = OUTPUT_FOLDER
```
RoyIsland, line 3: this need to be 3 folders back and not 2

ghost (approved):

bentheiii (approved):

PR #291, opened 2020-04-20
isolation after detection with probability

michaeleh:


src/corona_hakab_model/consts.py
```diff
@@ -136,6 +136,10 @@ class Consts(NamedTuple):
             ]),
     ]
     should_isolate_positive_detected: bool = False
+    isolate_after_num_day: int = 1  # will be in isolation the next day
+    p_will_obey_isolation: float = 1.0  # 100% will obey the isolation
+    isolation_factor: float = 0.0  # reduce 100%, meaning mult by 0
```
bgmoshe: The comment should be more descriptive, please state that this reduces the strength of the connections

src/corona_hakab_model/manager.py
```diff
@@ -137,23 +136,32 @@ def progress_tests_and_isolation(self, new_tests: List[PendingTestResult]):
...
         # TODO: Remove healthy agents from isolation?
+        self.isolate_agents()
 
         for new_test in new_tests:
             new_test.agent.set_test_start()
             self.pending_test_results.append(new_test)
 
+    def isolate_agents(self):
+        # those who had the detection test before isolate_after_num_day days exactly
+        can_be_isolated = self.current_step - self.date_of_last_test == self.consts.isolate_after_num_day
```
bgmoshe: please use brackets to avoid confusion

src/corona_hakab_model/consts.py
```diff
@@ -136,6 +136,10 @@ class Consts(NamedTuple):
             ]),
     ]
     should_isolate_positive_detected: bool = False
+    isolate_after_num_day: int = 1  # will be in isolation the next day
```
ghost: add those variables changes to simulation_parameters_example.py

src/corona_hakab_model/manager.py
```diff
@@ -137,23 +136,33 @@ def progress_tests_and_isolation(self, new_tests: List[PendingTestResult]):
...
+        # those who had the detection test before isolate_after_num_day days exactly
+        can_be_isolated = (self.current_step - self.date_of_last_test) == self.consts.isolate_after_num_day
+        # last test was positive and was the first one
+        tested_positive_once = self.tested_positive_vector & (self.tested_positive_vector_counter == 1)
+        remaining = np.array(self.agents)[can_be_isolated & tested_positive_once]  # get those agents
+        num_of_obedients = round(self.consts.p_will_obey_isolation * len(remaining))  # get number of agents to sample
+        will_obey_isolation = np.random.choice(remaining, num_of_obedients, replace=False)  # sample those who will obey
+
+        for agent in will_obey_isolation:
+            self.agents_in_isolation[agent.index] = True  # keep track about who is in isolation
```
bgmoshe, line 181: If I got sick but got healthy.died I am not isolated anymore

src/corona_hakab_model/manager.py
```diff
@@ -137,23 +136,33 @@ def progress_tests_and_isolation(self, new_tests: List[PendingTestResult]):
...
+            self.update_matrix_manager.change_agent_relations_by_factor(agent,
```
meirhalachmi, line 182: What will happen if the matrix will be reset? (e.g., `reset_policies_by_connection_type` will be called).
In that case, I think that this factor will be overridden.

We need to keep track of the isolated agents in the `UpdateMatrix` and prevent the reset method to effectively cancel their isolation

meirhalachmi (changes requested):

src/corona_hakab_model/manager.py
michaeleh: fixed

bgmoshe (approved):

meirhalachmi (approved):

PR #292, opened 2020-04-20
made all states possibly contagious

### LINKED ISSUE #286: Add an explicit contagiousness constant for each state
**POC: Meir Halachmi**

Right now, only 3 states have a non-zero contagiousness probability (A.K.A infection ratio):
image ratio of 0.

We need to add an explicit constant for each state, with a current default value of 0.

AmitGru:


src/corona_hakab_model/consts.py
```diff
@@ -302,27 +312,27 @@ class ImmuneTerminalState(ImmuneState, TerminalState):
...
             "NeedICU",
-            detectable=True,
+            contagiousness=self.need_icu_infection_ratio,
             test_willingness=self.need_icu_test_willingness
         )
 
-        improving_health = ImmuneStochasticState(
+        improving_health = ContagiousStochasticState(
             "ImprovingHealth",
-            detectable=True,
```
meirhalachmi, line 319: What happened to the detectability value of each state?

meirhalachmi (changes requested):

src/corona_hakab_model/consts.py
AmitGru: fixed

meirhalachmi (approved):

PR #300, opened 2020-04-20
extracted transition probabilities to consts.py

### LINKED ISSUE #293: Extract state machine transfer probabilities
Currently, all transfer probabilities between states are hard-coded inside the medical_state_machine function.
this way, you can't change these probabilities when creating a simulation from a json file.
this should be changed so that all probabilities will be taken from parameters located directly in the consts.py file

AmitGru:
Closes #293


src/corona_hakab_model/consts.py
```diff
@@ -76,6 +76,24 @@ class Consts(NamedTuple):
     # [10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35],
     # [0.013,0.016,0.025,0.035,0.045,0.053,0.061,0.065,0.069,0.069,0.065,0.063,0.058,0.053,0.056,0.041,0.040,0.033,
     # 0.030,0.025,0.020,0.015,0.015,0.015,0.010,0.010]))
+    # state machine transfer probabilities
+    # probability of '...' equals (1 - all other transfers)
+    # it should always come last after all other transition probabilities were defined
+    latent_to_asymptomatic_prob = 0.3
```
ghost: add those variables changes to simulation_parameters_example.py

tal-nc (approved):

src/corona_hakab_model/consts.py
AmitGru: done

src/corona_hakab_model/Parameters/simulation_parameters_example.py
```diff
@@ -46,6 +46,24 @@
     # [10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35],
     # [0.013,0.016,0.025,0.035,0.045,0.053,0.061,0.065,0.069,0.069,0.065,0.063,0.058,0.053,0.056,0.041,0.040,0.033,
     # 0.030,0.025,0.020,0.015,0.015,0.015,0.010,0.010]))
+    # state machine transfer probabilities
+    # probability of '...' equals (1 - all other transfers)
+    # it should always come last after all other transition probabilities were defined
+    latent_to_asymptomatic_prob = 0.3,
```
ghost: it needs to be in a dict format

AmitGru: oops, fixing it

src/corona_hakab_model/consts.py
```diff
@@ -76,6 +76,24 @@ class Consts(NamedTuple):
...
+    mild_to_pre_recovered_prob = ...
+    close_medical_care_to_icu_prob = 0.26
+    close_medical_care_to_mild_prob = ...
+    need_icu_to_deceased_prob = 0.0227
+    need_icu_to_improving_prob = ...
+    improving_to_need_icu_prob = 0.22
+    improving_to_pre_recovered_prob = 0.39
+    improving_to_mild_condition_prob = ...
+    pre_recovered_to_recovered_prob = ...
+    asymptomatic_to_recovered_prob = ...
```
ghost: you need to add type hint(i think it's called like that) in order for the json load to work
example
improving_to_pre_recovered_prob: int = 0.39

AmitGru: what's the type of '...'? also float?

ghost: i think it's type(...)

AmitGru: isn't that can cause problems when trying to give a float value to that parameter? e.g. if the user is trying to give all the probabilities in float and not using that option.

ghost: you're right, it should be Union[float, type(...)],
my bad

AmitGru: ok, cool

ghost (approved):

PR #308, opened 2020-04-21
Feature/policy affected circles to csv

### LINKED ISSUE #225: Shutdown information
We would like to recieve the following when making a shutdown on school/works -
- All schools, with their sizes, that were shut down.
- All schools, with their sizet, that remaind open.
- Same for workplaces.

erangold:
Added support for conditioned sampling in TabularSupervisable -- instead of receiving an interval, which dictates the sampling rate, the object is initialized with a generic predicate (which receives a SimulationManager object as input).
Added implementation of policy statistics supervisable -- upon policy activation, export statistics regarding the affected circles (their size, type, etc.)


src/corona_hakab_model/policies_manager.py
```diff
@@ -12,13 +12,17 @@ def __init__(self, manager: "SimulationManager"):
         self.update_matrix_manager = manager.update_matrix_manager
         self.consts = manager.consts
         self.logger = manager.logger
+        self.daily_affected_circles = None  # Dict(policy_id => Circle[])
```
bgmoshe: You can define it with type hinting :)

erangold: Just fixed it :)

src/corona_hakab_model/supervisor.py
```diff
@@ -258,27 +264,64 @@ def publish(self):
...
+        affected_circles_report = {
+            'policy_type': [], 'circle_size': [], 'circle_kind': [], 'circle_connection_type': []
+        }
+        for message, circles in policy_message_to_circles_dict.items():
+            self.update_affected_circles_report(affected_circles_report, circles, message)
+        return affected_circles_report
+
+    @staticmethod
+    def update_affected_circles_report(affected_circles_report, affected_circles, conditioned_policy_message):
+        affected_circles_report['policy_type'] += [conditioned_policy_message] * len(affected_circles)
```
bgmoshe: You can know the length of this from the start so why not use numpy array and then initialize it in vectorized form?

src/corona_hakab_model/supervisor.py
```diff
@@ -258,27 +264,64 @@ def publish(self):
...
+    def name(self) -> str:
+        return "policy_activation_stats"
+
+    def get(self, manager: "manager.SimulationManager"):
+        policy_message_to_circles_dict = manager.policy_manager.daily_affected_circles
+        affected_circles_report = {
+            'policy_type': [], 'circle_size': [], 'circle_kind': [], 'circle_connection_type': []
+        }
+        for message, circles in policy_message_to_circles_dict.items():
+            self.update_affected_circles_report(affected_circles_report, circles, message)
```
bgmoshe: I am surprised this compiles, its a static method you should call _AppliedPolicyReportSupervisable.method

src/corona_hakab_model/policies_manager.py
```diff
@@ -43,7 +48,18 @@ def activate_partial_opening_policies(self):
...
+                # Append affected circles to the daily affected circles
+                if activating_policy:
+                    self.update_daily_affected_circles(affected_circles, conditioned_policy)
+
+    def update_daily_affected_circles(self, affected_circles, conditioned_policy):
+        if self.daily_affected_circles is None:
+            self.daily_affected_circles = {}
+        if conditioned_policy.message not in self.daily_affected_circles:
+            self.daily_affected_circles[conditioned_policy.message] = []
+        self.daily_affected_circles[conditioned_policy.message] += affected_circles
```
bgmoshe: Why you are using the conditioned_policy message as key?

bgmoshe (changes requested):
Some questions

src/corona_hakab_model/supervisor.py
erangold: I fixed the function's call in a new commit

src/corona_hakab_model/policies_manager.py
erangold: I replaced the dictionary's key to be a conditioned policy (instead of its message)

src/corona_hakab_model/supervisor.py
erangold: I don't understand what you meant, can you elaborate?

bgmoshe: Oh wait, actually never mind this is a string

src/corona_hakab_model/main.py
```diff
@@ -157,6 +157,7 @@ def run_simulation(args):
             # Supervisable.GrowthFactor(
             #    Supervisable.Sum("Symptomatic", "Asymptomatic", "Latent", "Silent", "ICU", "Hospitalized"),
             Supervisable.CurrentInfectedTable(interval=consts.export_infected_agents_interval),
+            Supervisable.AppliedPolicyReportSupervisable(),
```
meirhalachmi: Comment it (but leave it in the code), It doesn't have to be a default

src/corona_hakab_model/policies_manager.py
```diff
@@ -9,13 +9,18 @@ def __init__(self, manager: "SimulationManager"):
         self.update_matrix_manager = manager.update_matrix_manager
         self.consts = manager.consts
         self.logger = manager.logger
+        # For each applied policy, save a list of affected circles
+        self.daily_affected_circles: Dict[ConditionedPolicy, List[SocialCircle]] = None
```
meirhalachmi: Wouldn't it be simpler to just initialize it to a `defaultdict(List)`, and empty it at the beginning of every `perform_policies` step?

meirhalachmi (changes requested):

src/corona_hakab_model/policies_manager.py
erangold: I wanted to separate the cases of "no policy was activated" and "some policies were activated, but none of the circles were affected". These two cases can't be separated if this variable contains an empty dictionary.

src/corona_hakab_model/policies_manager.py
```diff
@@ -9,13 +10,18 @@ def __init__(self, manager: "SimulationManager"):
         self.update_matrix_manager = manager.update_matrix_manager
         self.consts = manager.consts
         self.logger = manager.logger
+        # For each applied policy, save a list of affected circles
+        self.daily_affected_circles: Dict[ConditionedPolicy, List[SocialCircle]] = defaultdict(List)
```
bgmoshe, line 14: DefaultDict instead  of Dict

meirhalachmi (approved):

bgmoshe (approved):
Just fix type hinting to default dict

PR #333, opened 2020-04-24
matrix generation code clean-up

tal-nc:
some code clean-ups.
now also gets P as an input for work and school matrices creation.

this was originally made for (and merged into) dev-4.0, and is more mandatory for the mw matrix. However, since we are not going to use it any time soon, though it should be re-made for the current version.

### DIFF
--- a/src/corona_hakab_model/generation/matrix_generator.py
+++ b/src/corona_hakab_model/generation/matrix_generator.py
@@ -2,7 +2,7 @@ import logging
 import math
 from itertools import islice
 from random import random, choice, sample
-from typing import List
+from typing import List, Dict
 import os.path
 
 import bsa.universal
@@ -18,11 +18,10 @@ from generation.connection_types import (
     Geographic_Clustered_types,
     Random_Clustered_types,
 )
-from generation.matrix_consts import MatrixConsts
+from generation.matrix_consts import MatrixConsts, ConnectionTypeData
 from generation.node import Node
 from bsa.parasym import write_parasym, read_parasym
 from bsa.scipy_sparse import read_scipy_sparse
-import project_structure
 from project_structure import OUTPUT_FOLDER
 
 
@@ -30,7 +29,7 @@ class MatrixData:
     __slots__ = ("matrix_type", "matrix", "depth")
 
     # import/export variables
-    IMPORT_MATRIX_PATH = os.path.join(project_structure.OUTPUT_FOLDER, "matrix_data")
+    IMPORT_MATRIX_PATH = os.path.join(OUTPUT_FOLDER, "matrix_data")
 
     def __init__(self):
         self.matrix_type = None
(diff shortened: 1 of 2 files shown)


src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -172,194 +194,133 @@ def _create_random_clustered_circles_matrix(self, con_type: ConnectionTypes, cir
...
+                        possible_nodes = connected_nodes
 
-        # adding connections between all super small circles
-        self._add_small_circle_connections(super_small_circles_combined, connections, total_connections_float)
+                    # prevent connecting a connected node
+                    if len(possible_nodes) - len(node.connected) > 100:  # TODO 100 is made up. should be optimized
+                        possible_list = list(possible_nodes)
+                        random_friend = choice(possible_list)
+                        while random_friend in node.connected:
+                            random_friend = choice(possible_list)
```
RoyIsland, line 240: i am not sure i understood the reason for this

ghost, line 240: are we sure that this method works better than difference?

tal-nc: I timed it's runtime when P = 0, with and without this change, and it was significantly improved.
Doing difference update on large sets is quite expensive.
However, the switching condition is made up, and could probably be optimized. (not sure it's worth the trouble)

ghost, line 240: is there a chance that all the nodes are connected and this will run forever?

tal-nc: We can easily get to a point when possible nodes contains tens of thousands of nodes, or even more, while node.connected only contains a few. on those cases it is a very significant change.
I timed the matrix generation stage with population size of 20_000 and P = 0, and this change cut the run time by half. Should be even more significant as the population size increases.

tal-nc: No, that's why we have the if statement above.
this method is only faster if the amount of possible nodes is significantly higher than the connected amount, and otherwise the prior method is used.
i am sure that the condition could be optimized, however i think this should be good enough.

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -172,194 +194,133 @@ def _create_random_clustered_circles_matrix(self, con_type: ConnectionTypes, cir
...
+                    else:
+                        possible_nodes = possible_nodes.difference(node.connected).difference({first_connection})
+                        random_friend = choice(list(possible_nodes))
```
ghost, line 243: is there any chance that possible_nodes is gonna be empty?

tal-nc: Shouldn't be. That's why it now begins with initiating more than 1 node (was this way at work and school already, just added it to other too).

ghost (approved):

PR #334, opened 2020-04-25
analysis folders cleanup

tal-nc:
removed unnecessary folder generation analysis and moved population analysis into analyzers.
removed duplicated node.py file.
added __init__.py for generation


src/corona_hakab_model/analyzers/population_analysis.py
```diff
@@ -90,7 +90,7 @@ def export_csv(self, circle_file_path: str, age_file_path: str):
     parser.add_argument("--no-view",dest='view',default=True,action='store_false',help="Skip viewing the generated histograms.")
     parser.add_argument("--no-save",dest='save',default=True,action='store_false',help="Skip saving the resulting CSV")
     args = parser.parse_args()
-    directory = Path(args.directory or project_structure.OUTPUT_FOLDER)
+    directory = Path(args.directory or project_structure.MODEL_FOLDER / "output")
```
bgmoshe, line 93: Why? Isn't it more appropriate to bundle the anlysis of the population with the population data in our output folder?

ghost (approved):

src/corona_hakab_model/generation/generation_manager.py
```diff
@@ -40,6 +40,6 @@ def export(self):
     def import_circles_and_create_matrices(self):
```
ghost: maybe remove those too?

README.md
```diff
@@ -106,7 +106,7 @@
 - To use run **python main.py shift-real-life** with first argument as csv to compare to, second argument simulation output
 ### Population Analyser
 - The population analyser reads a population data file (generated with the simulator, usually "population_data.pickle") and outputs a histogram of population ages, and social circles sizes by type.
-- to use, run **python generation_analysis\population_analysis.py**
+- to use, run **python analyzers\population_analysis.py**
 - Most commonly, you will use the **[-d|--directory]** option to specify the directory to read. The directory is expected to contain a file named "population_data.pickle"
 - You can also specify the input population data file using **[-p|--population]**, and output files using **--circle** and **--age**.
 - As with all runnables, additional help can be found by running **python generation_analysis\population_analysis.py --help**
```
ghost: change here to analyzers too

ghost (changes requested):
need a little change

bgmoshe (commented):
Small comment. Plus is node.py not  in use?

ghost:
> Small comment. Plus is node.py not in use?

It is in use, we had 2 of those. One in generation and one that tal deleted

src/corona_hakab_model/generation/generation_manager.py
tal-nc: done

README.md
tal-nc: done

src/corona_hakab_model/analyzers/population_analysis.py
tal-nc: population analysis is not done in the same run as the rest of the simulation,it is done afterwards (maybe we should change that, but this is how it works right now).

OUTPUT_FOLDER was updated to be:
`OUTPUT_FOLDER = MODEL_FOLDER / "output" / output_subdir
which means, it is looking for a subdir for the current run. since the population analysis is done alone, there is no such folder. (it is looking for a folder inside outputs, with a named based on the current time, and since this folder doesn't exists, it crushed).

using MODEL_FOLDER / "output" is the only way to get to the output folder right now. maybe we should change that too.

ghost (approved):

PR #336, opened 2020-04-25
Issue 335: Added number of days to get test result as a distribution …

### LINKED ISSUE #335: Change number of days to get test result as a distribution
Currently, we have a constant number indicating number of days to get the results. This does not reflect the truth, we want to convert this to a distribution (currently constant distribution)

bgmoshe:
I am not sure if we should make it a class that gets as input the kind of distribution. If we will have more distributions that are not uniform, I'll change that.

### DIFF
--- a/src/corona_hakab_model/Parameters/simulation_parameters_example.py
+++ b/src/corona_hakab_model/Parameters/simulation_parameters_example.py
@@ -1,3 +1,5 @@
+from util import get_numpy_uniform_dist
+
 {
     # medical states
     "LATENT": "Latent",
@@ -131,7 +133,7 @@
                 "AsymptomaticEnd": .98,
                 "Latent-Asymp": .98,
                 "Latent-Presymp": .98
-            }, time_until_result=3),
+            }, time_dist_until_result=get_numpy_uniform_dist(3)),  # Constant distribution
             daily_num_of_tests_schedule={0: 100, 10: 1000, 20: 2000,
                                          50: 5000},
             testing_gap_after_positive_test=2,
@@ -163,7 +165,7 @@
                 "AsymptomaticEnd": .92,
                 "Latent-Asymp": .92,
                 "Latent-Presymp": .92
-            }, time_until_result=5),
+            }, time_dist_until_result=get_numpy_uniform_dist(5)),  # Constant distribution
             daily_num_of_tests_schedule={0: 500, 10: 1500, 20: 2500,
                                          50: 7000},
             testing_gap_after_positive_test=3,
--- a/src/corona_hakab_model/Parameters/simulation_parameters_example.py
+++ b/src/corona_hakab_model/Parameters/simulation_parameters_example.py
@@ -1,5 +1,3 @@
-from util import get_numpy_uniform_dist
-
 {
     # medical states
     "LATENT": "Latent",
--- a/src/corona_hakab_model/consts.py
+++ b/src/corona_hakab_model/consts.py
@@ -12,7 +12,7 @@ from medical_state import ContagiousState, ImmuneState, MedicalState, Susceptibl
 from medical_state_machine import MedicalStateMachine
 from policies_manager import ConditionedPolicy, Policy
 from state_machine import StochasticState, TerminalState
-from util import dist, rv_discrete, upper_bound, BucketDict
+from util import dist, rv_discrete, upper_bound, BucketDict, get_numpy_uniform_dist
 
 """
 Overview:
@@ -160,7 +160,7 @@ class Consts(NamedTuple):
                 ASYMPTOMATIC_END: .98,
                 LATENT_ASYMP: .98,
                 LATENT_PRESYMP: .98
-            }, time_until_result=3),
+            }, time_dist_until_result=get_numpy_uniform_dist(3)),  # Constant distribution
             daily_num_of_tests_schedule={0: 100, 10: 1000, 20: 2000, 50: 5000},
             testing_gap_after_positive_test=2,
             testing_gap_after_negative_test=1,
@@ -190,7 +190,7 @@ class Consts(NamedTuple):
                 ASYMPTOMATIC_END: .92,
                 LATENT_ASYMP: .92,
                 LATENT_PRESYMP: .92
-            }, time_until_result=5),
+            }, time_dist_until_result=get_numpy_uniform_dist(5)),  # Constant distribution
             daily_num_of_tests_schedule={0: 500, 10: 1500, 20: 2500, 50: 7000},
             testing_gap_after_positive_test=3,
             testing_gap_after_negative_test=1,
--- a/src/corona_hakab_model/consts.py
+++ b/src/corona_hakab_model/consts.py
@@ -289,6 +289,7 @@ class Consts(NamedTuple):
             "np": np,
             "BucketDict": BucketDict,
             "len": len,
+            "get_numpy_uniform_dist": get_numpy_uniform_dist
         }
 
         parameters = eval(data, expressions)
--- a/src/corona_hakab_model/detection_model/healthcare.py
+++ b/src/corona_hakab_model/detection_model/healthcare.py
@@ -35,8 +35,6 @@ class DetectionTest:
         self.time_dist_until_result = time_dist_until_result
 
     def get_times_to_get_results(self, number_of_agents=1):
-        if number_of_agents == 1:
-            return self.time_dist_until_result()
         return self.time_dist_until_result(size=number_of_agents)
 
     def test(self, agent: Agent):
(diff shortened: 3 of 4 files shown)


src/corona_hakab_model/util.py
```diff
@@ -32,6 +32,14 @@ def upper_bound(d):
     return d.b + d.kwds.get("loc", 0)
 
 
+def get_numpy_uniform_dist(u_bound, l_bound=None):
```
ghost: switch l_bound and u_bound position

src/corona_hakab_model/consts.py
```diff
@@ -190,7 +190,7 @@ class Consts(NamedTuple):
                 ASYMPTOMATIC_END: .92,
                 LATENT_ASYMP: .92,
                 LATENT_PRESYMP: .92
-            }, time_until_result=5),
+            }, time_dist_until_result=get_numpy_uniform_dist(5)),  # Constant distribution
```
ghost, line 193: add to consts_parameters_example

ghost, line 193: add get_numpy_uniform_dist in eval that is in from_json func

ghost (changes requested):
small changes

src/corona_hakab_model/detection_model/healthcare.py
```diff
@@ -28,14 +28,24 @@ class PendingTestResults(Queue[PendingTestResult]):
...
 class DetectionTest:
-    def __init__(self, state_to_detection_prop: Dict[MedicalState, float], time_until_result):
+    def __init__(self,
+                 state_to_detection_prop: Dict[MedicalState, float],
+                 time_dist_until_result):
         self.state_to_detection_prop = deepcopy(state_to_detection_prop)
-        self.time_until_result = time_until_result
+        self.time_dist_until_result = time_dist_until_result
+
+    def get_times_to_get_results(self, number_of_agents=None):
```
tal-nc: why not set number_of_agents to 1 as default, and get rid of the if statement?

src/corona_hakab_model/util.py
bgmoshe: Why? This makes less sense in my mind, think that if we give ONE parameter we want it to be the upper bound and not the lower bound

ghost: but than you'll have to use the function like so get_numpy_uniform_dist(5, 3)

ghost: i think this will be better

def get_numpy_uniform_dist(l_bound, u_bound=None):
    if u_bound:
        range_to_choose_from = list(range(l_bound, u_bound))
    else:
        range_to_choose_from = [l_bound]
    return lambda size=None: np.random.choice(range_to_choose_from, size=size)

src/corona_hakab_model/detection_model/healthcare.py
```diff
@@ -28,14 +28,24 @@ class PendingTestResults(Queue[PendingTestResult]):
...
-    def __init__(self, state_to_detection_prop: Dict[MedicalState, float], time_until_result):
+    def __init__(self,
+                 state_to_detection_prop: Dict[MedicalState, float],
+                 time_dist_until_result):
         self.state_to_detection_prop = deepcopy(state_to_detection_prop)
-        self.time_until_result = time_until_result
+        self.time_dist_until_result = time_dist_until_result
+
+    def get_times_to_get_results(self, number_of_agents=1):
+        if number_of_agents == 1:
```
ghost, line 38: no need for the if, you can just pass number_of_agents

src/corona_hakab_model/Parameters/simulation_parameters_example.py
```diff
@@ -1,3 +1,5 @@
+from util import get_numpy_uniform_dist
```
ghost: no need to import

src/corona_hakab_model/util.py
bgmoshe: I thought about this for a bit b4 ur comment, can u check my new implementation. I get what you say, it was just weird to implement like it

ghost: hmmm it's a bit complicated for this problem, why is it better than my solution?

src/corona_hakab_model/detection_model/healthcare.py
bgmoshe: I now rememeber why I did this. This function should return a number when size=1,  but if you give choice 1 it return an ndarray, so if I give 1 as input I call it without size=1 (look at the implementation)

src/corona_hakab_model/Parameters/simulation_parameters_example.py
bgmoshe: I think this all file is a joke, we should remove all this files

src/corona_hakab_model/util.py
```diff
@@ -32,6 +32,17 @@ def upper_bound(d):
     return d.b + d.kwds.get("loc", 0)
 
 
+def get_numpy_uniform_dist(l_bound=None, u_bound=None):
```
ghost: i think you should not allow l_bound to be None because it must have a value either way

ghost: sorry for all the comments... i know i'm annoying

src/corona_hakab_model/Parameters/simulation_parameters_example.py
ghost: i agree... but the problem is that tom and shir are using the config files.
so it's easier for them to copy new values from this file

src/corona_hakab_model/util.py
```diff
@@ -32,6 +32,17 @@ def upper_bound(d):
...
+    assert l_bound is not None and u_bound is None\
```
tal-nc: why asserting u_bound is None?

i see that the variables names are confusing because sometimes the first variable is lower bound, and some times it is the higher bound. 
why not just give them generic names ...(a, b=None ).
this way you enforce at least one value without the assert, and the names don't get confused

bgmoshe: I somewhat agree ( I do not like a and b cause they are not such meaningfull name, instead I added a doc)

src/corona_hakab_model/util.py
bgmoshe: I thought about furhet extending this (to not giving a but do giving b)
I decieded it is not worth the effort, so done :)

src/corona_hakab_model/Parameters/simulation_parameters_example.py
bgmoshe: I get it, but maybe we should auto-generate it instead of filling everytime (i.e. having a command/scrpit that generates them)

bgmoshe: Also it is called .py that why the import was needed, it should be .json

src/corona_hakab_model/util.py
```diff
@@ -32,14 +32,17 @@ def upper_bound(d):
...
-    assert l_bound is not None and u_bound is None\
-        , "Uniform distribution must have at least one parameter."
-
-    if u_bound:
-        range_to_choose_from = list(range(l_bound, u_bound))
+def get_numpy_uniform_dist(a, b=None):
+    """
+    :param a: lower bound of an interval
+    :param b: upper bound if the interval. If b=None than we have discrete distribution.
+    :return: Return a distribution that gets number of elements to sample. If b=0 this is a discrete distribution.
```
RoyIsland, line 39: if b=None

src/corona_hakab_model/util.py
```diff
@@ -32,14 +32,17 @@ def upper_bound(d):
...
-def get_numpy_uniform_dist(l_bound=None, u_bound=None):
-    assert l_bound is not None and u_bound is None\
-        , "Uniform distribution must have at least one parameter."
-
-    if u_bound:
-        range_to_choose_from = list(range(l_bound, u_bound))
+def get_numpy_uniform_dist(a, b=None):
+    """
+    :param a: lower bound of an interval
+    :param b: upper bound if the interval. If b=None than we have discrete distribution.
```
ghost:
```suggestion
    :param b: upper bound of the interval. If b=None than we have discrete distribution.
```

src/corona_hakab_model/util.py
ghost:
```suggestion
    :return: Return a distribution that gets number of elements to sample. If b=None this is a discrete distribution.
```

src/corona_hakab_model/Parameters/simulation_parameters_example.py
ghost: I agree that generating them and making them be json is the correct approach

ghost (approved):

PR #338, opened 2020-04-25
added random connection in geo

### LINKED ISSUE #337: Add geographic random connections

ghost:


src/corona_hakab_model/infection.py
```diff
@@ -63,11 +63,28 @@ def _perform_infection(self) -> Dict[Agent, InfectionInfo]:
 
     def _infect_random_connections(self) -> Dict[Agent, InfectionInfo]:
         connections = self.manager.num_of_random_connections * self.manager.random_connections_factor
-        probs_not_infected_from_connection = np.ones_like(connections, dtype=float)
+
+        probs_not_infected_from_random_connection = self._get_probs_not_infected_from_random_connection(connections)
+        probs_not_infected_from_geo_random_connection = self._get_probs_not_infected_from_geo_random_connection(connections)
+
+        probs_not_infected_from_connection = (probs_not_infected_from_random_connection + probs_not_infected_from_geo_random_connection) - 1
```
tal-nc, line 74: why the -1 at the end?

src/corona_hakab_model/infection.py
```diff
@@ -121,9 +155,22 @@ def _get_random_infection_info(self, agent_id: int, infection_probs: np.ndarray)
...
+
+        if connection_type in connection_types.With_Geo_Random_Connections:
+            agent_connection_type_circle = [circle for circle in self.manager.social_circles_by_agent_index[agent_id]
+                                            if circle.connection_type == connection_type][0]
+            agent_geo_circle = [geo_circle for geo_circle in self.manager.geographic_circles
+                                for circle in geo_circle.connection_type_to_social_circles[connection_type]
+                                if circle.guid == agent_connection_type_circle.guid][0]
+            infectious_circles = \
+                [circle for circle in agent_geo_circle.connection_type_to_social_circles[connection_type]]
+            agents_id = [agent.index for circle in infectious_circles for agent in circle.agents]
```
tal-nc, line 177: i am not sure i understand why you did it this way?
why not use:
```
geo_circle = manager.geographic_circle_by_agent_index[agent_id]
agents_id = [agent.index for circle in geo_circle.connection_type_to_social_circles[connection_type] for agent in circle.agents]
```

tal-nc (changes requested):

src/corona_hakab_model/infection.py
ghost: because the values should be between 0-1

src/corona_hakab_model/infection.py
ghost: i did that in the beginning and understood that this is not correct, the agent can work in a social circle that is not in his own geo circle.
took my ages to find and solve this problem.

tal-nc: good point.
won't the following be good enough?
`agent_geo_circle = [geo_circle for geo_circle in self.manager.geographic_circles 
                                 if agent_connection_type_circle in geo_circle.connection_type_to_social_circles[connection_type]][0]`

default comparison should be good enough here, and i don't think there is any reason to override it by comparing the .guid of the circles.

tal-nc (approved):
wrote another small suggestion, but overall looks great

ghost:
> wrote another small suggestion, but overall looks great

i agree, i'll change it

src/corona_hakab_model/infection.py
ghost: done

src/corona_hakab_model/infection.py
```diff
@@ -80,18 +97,35 @@ def _infect_random_connections(self) -> Dict[Agent, InfectionInfo]:
...
+                if len(agents_id) == 1:
+                    # One-Agent in circles, you can't randomly meet yourself..
+                    continue
+
+                total_infectious_random_connections = np.dot(
+                    self.manager.contagiousness_vector[agents_id],
+                    connections[agents_id, connection_type],
+                )
+
+                circles_total_random_connections = sum([circle.total_random_connections for circle in circles])
```
meirhalachmi: use np.sum, it's much faster

src/corona_hakab_model/infection.py
meirhalachmi, line 74: kinda hacky but ok :laughing:

src/corona_hakab_model/infection.py
```diff
@@ -121,9 +155,20 @@ def _get_random_infection_info(self, agent_id: int, infection_probs: np.ndarray)
...
-            [c for c in self.manager.social_circles_by_agent_index[agent_id] if c.connection_type == connection_type][0]
-        agents_id = [a.index for a in infectious_circle.agents]
+        agents_id = []
+
+        if connection_type in connection_types.With_Random_Connections:
+            infectious_circle = next(circle for circle in self.manager.social_circles_by_agent_index[agent_id]
+                                     if circle.connection_type == connection_type)
+            agents_id = [agent.index for agent in infectious_circle.agents]
+
+        if connection_type in connection_types.With_Geo_Random_Connections:
```
meirhalachmi, line 169: There are (and probably will always be) _O(1)_ geo-random connection type and _O(1)_ geo-circles, hence _O(1)_ possible lists of agents_id.
However, this list is constructed again and again each time this function is called and the connection_type=WORK is chosen. 

Maybe you should store a dictionary of [all the people with circle type Y in geo-circle X] to allow quick access to this data. (There are many such dictionaries in the manager \ circ generator)

meirhalachmi (changes requested):

src/corona_hakab_model/infection.py
ghost: is there a better solution? i couldn't really think of one :sweat_smile:

meirhalachmi (approved):

PR #339, opened 2020-04-25
added automatic export to consts as a json file

### LINKED ISSUE #296: Add export_to_file method to consts files
We would like to be able to export a consts object. It could be exported either as .json file or as .py file.
Note that this functionality should be added to each of the consts files (consts, circles_consts and matrix_consts).

tal-nc:
we need a better export to the consts files than we currently have (hard-coping the parameters files).
this should do, however i am not sure that it is perfect either. the picklejson extracts a bit too much information. mainly rv_discretes are VARY annoying in this aspect.
i am not using regular json export since it can't handle complex objects.
the matrix_consts and circles_consts are great, but the simulation_consts comes out a bit too long.

note that this also adds a dependency to the project, i added it to the requirements.txt, let me know if it needs to be added elsewhere.

i think this is good enough, but i would like to hear your opinion. 

we also need to create a base class for all the consts, but that's for another conversation.

### DIFF
--- a/src/corona_hakab_model/generation/generation_manager.py
+++ b/src/corona_hakab_model/generation/generation_manager.py
@@ -16,7 +16,7 @@ class GenerationManger:
     generation manager can export the entire generation information as a json.
     """
 
-    __slots__ = ("matrix_data", "population_data")
+    __slots__ = ("matrix_data", "population_data", "circles_consts", "matrix_consts")
 
     def __init__(self, circles_consts: CirclesConsts, matrix_consts: MatrixConsts):
         # setting logger
@@ -32,6 +32,12 @@ class GenerationManger:
         matrix_generation = MatrixGenerator(circles_generation.population_data, matrix_consts=matrix_consts)
         self.matrix_data = matrix_generation.matrix_data
 
+        # save consts to allow export
+        self.circles_consts = circles_consts
+        self.matrix_consts = matrix_consts
+
     def save_to_folder(self, folder):
         self.matrix_data.export(os.path.join(folder, 'matrix_data'))
         self.population_data.export(folder, 'population_data')
+        self.circles_consts.export(folder, 'circles_consts.json')
+        self.matrix_consts.export(folder, 'matrix_consts.json')
--- a/src/corona_hakab_model/generation/matrix_consts.py
+++ b/src/corona_hakab_model/generation/matrix_consts.py
@@ -1,4 +1,6 @@
+import os
 from typing import Dict, NamedTuple
+import jsonpickle
 import numpy as np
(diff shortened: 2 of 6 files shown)


bgmoshe:
@tal-nc Can you attach a result json file for example?

tal-nc:
[consts-jsons.zip](https://github.com/CoronaHakab/CoronaHakabModel/files/4534923/consts-jsons.zip)

ghost:
I think it's a great solution, we just need to switch rv_discrete to choice and it will look much better.
I think that we should switch our ugly format with this and add an option to import this format.

bgmoshe:
Is there a way to get rid of :
image

tal-nc:
> Is there a way to get rid of :
> image

not that i know of

src/corona_hakab_model/generation/generation_manager.py
```diff
@@ -32,6 +32,12 @@ def __init__(self, circles_consts: CirclesConsts, matrix_consts: MatrixConsts):
...
         self.matrix_data = matrix_generation.matrix_data
 
+        # save consts to allow export
+        self.circles_consts = circles_consts
+        self.matrix_consts = matrix_consts
+
     def save_to_folder(self, folder):
         self.matrix_data.export(os.path.join(folder, 'matrix_data'))
         self.population_data.export(folder, 'population_data')
+        self.circles_consts.export(folder, 'circles_consts.json')
```
bgmoshe, line 42: Wouldn't it be more appropriate to have export function for each of them instead of bundling them here?

src/corona_hakab_model/generation/matrix_consts.py
```diff
@@ -86,6 +88,12 @@ def from_file(cls, param_path):
 
         return cls(**parameters)
 
+    def export(self, export_path, file_name: str = "matrix_consts.json"):
+        if not file_name.endswith(".json"):
+            file_name += ".json"
+        with open(os.path.join(export_path, file_name), "w") as export_file:
+            export_file.write(jsonpickle.encode(self._asdict()))
```
bgmoshe, line 95: doesnt dict(self) works?

bgmoshe (approved):
Overall looks good, just some small question

src/corona_hakab_model/generation/matrix_consts.py
tal-nc: no, returns an error. 

we can however do: 
`jsonpickle.endoce(self)`
and encode the namedtuple directly. it would probably be easier to decode straight into a consts object, but it is less readable as a json

bgmoshe (approved):

PR #342, opened 2020-04-26
code cleanup - common folder

tal-nc:
there are way too many files directly in corona_hakab_model. i think we should organize it.
started by adding 'common' folder for commonly used code.
also split a few multi-objects files into different files (no reason for Circle to be in agent.py).


src/corona_hakab_model/common/agent.py
```diff
@@ -6,12 +6,12 @@
 import pandas as pd
 from numpy import nan
 
-from generation.circles import SocialCircleConstraint
+from common.social_circle import SocialCircleConstraint
```
ghost:
```suggestion
from .social_circle import SocialCircleConstraint
```

ghost:
Maybe move detection_model/detection_testing_types.py too?

src/corona_hakab_model/common/__init__.py
```diff
@@ -0,0 +1,10 @@
+__all__ = [
```
ghost: why do we need this?

ghost (changes requested):
I really like this change, it will make the project look a lot better.
Just some small fixes and we'll merge it.

src/corona_hakab_model/common/agent.py
tal-nc: done

src/corona_hakab_model/common/__init__.py
tal-nc: i though we might wanna use it some times, it allows us to do 
`from common import *`
however we are not using it now, so i will remove it

ghost (approved):
looks good, good job

src/corona_hakab_model/common/__init__.py
ghost: oh that's what it does, i didn't know

ghost:
just need to fix a little conflict

PR #351, opened 2020-04-29
Track close connections

### LINKED ISSUE #340: Track close connections
In order to isolate the close circle of any detected patient, we need to (initially) store the close connections of each agent.
This information exists in the matrix generation phase.
We need to find an elegant way to store this information in the generation phase and use it in the simulation.

In addition, since the generation can be done separately from the simulation, we probably need to store this info in the PopulationData and pickle it as well.

denzo9:


bgmoshe:
@denzo9  No issue is linked, can you open one describing what you were asked to do and link it to here? I cannot review this if I am not sure what you were asked to do

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -152,28 +184,42 @@ def _add_layer(self, con_type_data: ConnectionTypeData, connections: List[List[i
...
+                    strengthes[np.where(conns == conn)] = known_strength
                     del known_strengths[(conn, agent.index)]
                 else:
                     # connection is new. store the strength for future use
-                    known_strengths[(agent.index, conn)] = strengthes[np.where(conns==conn)]
+                    known_strengths[(agent.index, conn)] = strengthes[np.where(conns == conn)]
+
+            daily_connections = [conn for conn in conns if strengthes[np.where(conns == conn)] ==
+                                 con_type_data.connection_strength]
+            weekly_connections = [conn for conn in conns if strengthes[np.where(conns == conn)] !=
```
bgmoshe: If it's daily_connections 's complement, then:
* if conn not in daily_connections  would be more appropriate, no?

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -24,6 +25,36 @@
...
+class ConnectionData:
+    __slots__ = (
+        "connected_ids_by_strength",
+    )
+
+    def __init__(self, agents):
+        self.connected_ids_by_strength = {agent: {connection_type: AgentConnections() for connection_type
+                                                                      in ConnectionTypes} for agent in agents}
+
+    def export(self, export_path, file_name: str):
```
tal-nc: add a default name

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -24,6 +25,36 @@
 from project_structure import OUTPUT_FOLDER
 
 
+class AgentConnections(NamedTuple):
+    daily_connections: List = set()
+    weekly_connections: List = set()
```
tal-nc: type hinting doesn't match the used type.
change type hinting to Set.

meirhalachmi: Did you assume my type?!?!?

denzo9: oh right, changed one but not the other

src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -24,6 +25,36 @@
 from project_structure import OUTPUT_FOLDER
 
 
+class AgentConnections(NamedTuple):
+    daily_connections: set = set()
```
bgmoshe: Use the correct type, i.e. Set (import it from Typing)
Also, it would be super if you can use Set[float] isntead to further hint about this variable

denzo9: Wandered why it corrected me from Set to set... 
Changed it to Set[int]

bgmoshe (approved):
Looks OK.

PR #353, opened 2020-04-29
reset_current_limitations or not

meirhalachmi:
+ there is no reason to set any value other than "False" in "policy.active"

@bgmoshe, please make sure that it doesn't conflict with your current work. 
If it's ok, I'll un-draft this PR.

### DIFF
--- a/src/corona_hakab_model/policies_manager.py
+++ b/src/corona_hakab_model/policies_manager.py
@@ -113,15 +113,17 @@ class ConditionedPolicy:
     this class contains a policy that is supposed to run when a given condition is satisfied.
     """
 
-    __slots__ = "activating_condition", "policy", "active", "reset_current_limitations", "message"
+    __slots__ = "activating_condition", "policy", "active", "dont_repeat_while_active", "reset_current_limitations", "message"
 
     def __init__(
             self, activating_condition: Callable[[Any], bool], policy: Policy,
             reset_current_limitations=True,
+            dont_repeat_while_active=True,
             message=""
     ):
-        self.reset_current_limitations = reset_current_limitations
         self.activating_condition = activating_condition
         self.policy = policy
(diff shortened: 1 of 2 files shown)


tal-nc:
sometimes it is useful to initiate a policy as actived. 
it is used for policies that their condition is satisfied from the get go, but you don't want them to run before another policy happens.
for example, i once ran with 2 policies: when there are more than X currently sick agents close all schools, when there are less than Y open them. 
in such a case, initiating the 2nd policy as active saves some time.

meirhalachmi:
@tal-nc I understand what you're saying, but I don't think that the `active` flag is the right way to get this behavior. Especially since any other activated policy will switch this policy's `active` flag to `False`, and generally, that's not the meaning of "is the policy active".

I think we should try to look for a better activation condition mechanism that would be better suited to handle such cases (e.g., a condition that maintains the history of the indication which it examines, and returns `True` only if the current number of sick agents is both less than Y and lower than the previous number of sick agents)

ghost (approved):

src/corona_hakab_model/policies_manager.py
```diff
@@ -113,10 +113,18 @@ class ConditionedPolicy:
...
 
-    __slots__ = "activating_condition", "policy", "active", "message"
+    __slots__ = "activating_condition", "policy", "active", "dont_repeat_while_active", "reset_current_limitations", "message"
 
-    def __init__(self, activating_condition: Callable[[Any], bool], policy: Policy, active=False, message=""):
+    def __init__(
+            self, activating_condition: Callable[[Any], bool], policy: Policy,
+            reset_current_limitations=True,
+            dont_repeat_while_active=True,
+            active=True,
```
tal-nc: maybe set default to false tho

meirhalachmi: Oh crap

PR #358, opened 2020-05-01
added config copy

ghost:
…t files

### DIFF
--- a/linux/gen_and_run_parallel.sh
+++ b/linux/gen_and_run_parallel.sh
@@ -3,9 +3,14 @@ num_of_runs=$1
 circle_consts=$2
 matrix_consts=$3
 simulation_consts=$4
-output_folder=$5
+TIMESTAMP=`date --utc -d "+3 hours" +_%d%m%Y_%H`
+output_folder=$5$TIMESTAMP
+
 
 python3.8 ./src/corona_hakab_model/main.py generate -c $circle_consts -m $matrix_consts -o $output_folder
+cp $circle_consts $output_folder
+cp $matrix_consts $output_folder
+cp $simulation_consts $output_folder
 
 
 for ((i=0; i< $num_of_runs - 1; i++))
--- a/linux/gen_and_run_parallel.sh
+++ b/linux/gen_and_run_parallel.sh
@@ -3,7 +3,7 @@ num_of_runs=$1
 circle_consts=$2
 matrix_consts=$3
 simulation_consts=$4
-TIMESTAMP=`date --utc -d "+3 hours" +_%d%m%Y_%H`
+TIMESTAMP=`date --utc -d "+3 hours" +_%d%m%Y_%H_%M`
 output_folder=$5$TIMESTAMP
 
 


linux/gen_and_run_parallel.sh
```diff
@@ -3,9 +3,14 @@ num_of_runs=$1
 circle_consts=$2
 matrix_consts=$3
 simulation_consts=$4
-output_folder=$5
+TIMESTAMP=`date --utc -d "+3 hours" +_%d%m%Y_%H`
+output_folder=$5$TIMESTAMP
```
bgmoshe, line 7: Why not use minutes and seconds also?

linux/gen_and_run_parallel.sh
```diff
@@ -3,9 +3,14 @@ num_of_runs=$1
 circle_consts=$2
 matrix_consts=$3
 simulation_consts=$4
-output_folder=$5
+TIMESTAMP=`date --utc -d "+3 hours" +_%d%m%Y_%H`
```
bgmoshe: Why do you use --utc and +3 ?
That is incorrect, Israel is GMT +2 (but sometime can also be +3)
Why not just use date +_%d%m%Y_%H

bgmoshe (changes requested):
Look overall good, just few fixes

linux/gen_and_run_parallel.sh
ghost: because the time will be wrong if we don't add the 3 hours, you are correct about the clock change but i don't think we can solve it easily

linux/gen_and_run_parallel.sh
ghost: i don't think we really need the minutes and seconds, we only added the hour when we did it manually

linux/gen_and_run_parallel.sh
bgmoshe: You can, do not use --utc

linux/gen_and_run_parallel.sh
bgmoshe: get what you say, but in the future it might cause a problem, so you should check if the dir exists or try to make it more unique. Maybe seconds is exgeration, but I think minutes is mandatory

bgmoshe:
@RoyIsland  Also, have you seen what Idan wrote about:
aws s3 cp <input file or dir> s3://omni-docker/<output path>

linux/gen_and_run_parallel.sh
ghost: the default time is the machine time, which is not israel's time

ghost:
> @RoyIsland Also, have you seen what Idan wrote about:
> aws s3 cp s3://omni-docker/

yep but i dont really know how to use it, so for now i think we will keep using the storage in our machines

linux/gen_and_run_parallel.sh
ghost: done

linux/gen_and_run_parallel.sh
bgmoshe: Mhmmm I dont like this so much, but I guess it is OK for now :-/

bgmoshe (approved):

PR #360, opened 2020-05-01
added policy props set/update

meirhalachmi:


src/corona_hakab_model/policies_manager.py
```diff
@@ -78,15 +78,22 @@ def __init__(
             connection_change_factor: float,
             circle_conditions: Iterable[Callable[[Any], bool]],
             agent_conditions: Iterable[Callable[[Any], bool]] = None,
+            policy_props_update: Dict = None,
```
ghost, line 81:
```suggestion
            policy_props_update: Dict = dict(),
```

src/corona_hakab_model/policies_manager.py
```diff
@@ -78,15 +78,22 @@ def __init__(
...
         self.circle_conditions = circle_conditions
 
+        if policy_props_update is None:
+            policy_props_update = dict()
+
         if agent_conditions is None:
             agent_conditions = []
 
         self.agent_conditions = agent_conditions
+        self.policy_props_update = policy_props_update
```
ghost, line 95: you have this initialization twice

src/corona_hakab_model/policies_manager.py
meirhalachmi: giving a param a default mutable value is problematic:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

src/corona_hakab_model/policies_manager.py
meirhalachmi: oops

src/corona_hakab_model/policies_manager.py
ghost: oh i didn't know about that, sounds good

ghost (approved):

PR #362, opened 2020-05-03
generation bugfix

tal-nc:


bgmoshe:
It loos OK. How did you notice the bug and how did you check that this fixed it?

tal-nc:
added another small fix for odd numbers.
noticed there were differences between running with other connection only, vector alone vs the matrix alone.
looked into it a bit and i saw that the matrix generated too many connections. 
it lead me there. it should be (and used to be) divided by 2. it got broke recently. 

i am testing this by running on a custom population and comparing the matrix total to the expected one

src/corona_hakab_model/generation/matrix_consts.py
```diff
@@ -127,6 +127,22 @@ def get_rounded_connections_amount(self, shape: int) -> np.ndarray:
...
 
+    def get_scale_free_connections_amount(self, shape: int) -> np.ndarray:
+        """returns the total amount of connections for the given type
+        randomly chooses between floor and ceil such that the average will be correct
+        raises an error if daily or weekly connections amount is undefined"""
+
+        assert self.daily_connections_amount is not None and self.weekly_connections_amount is not None, \
+            "rolled daily or weekly connection on a type without daily or weekly connections"
+        total_connections = self.total_connections_amount / 2
+        if total_connections % 1 == 0:
```
meirhalachmi, line 138: That's the most creative method I've ever seen to check if a number is an integer or not.

meirhalachmi (approved):

PR #364, opened 2020-05-04
analysis tools

tal-nc:
added an analysis tools folder.
should contain tools used for analysis such as israel comparison.
created there a script that gets a list of simulations folders (each created from gen_and_run_parallel) and compares them

### DIFF
--- a/analysis-tools/multiple_runs_comparison.py
+++ b/analysis-tools/multiple_runs_comparison.py
@@ -36,15 +36,16 @@ class Simulation:
 
 # configuration:
 STATS_TO_COMPARE = ["was ever sick"]
-TITLE = "matrix vs vector no policy comparison"
-TO_SHOW = False
-TO_SAVE = True
+TITLE = "multiple matrix only comparisons"
+TO_SHOW = True
+TO_SAVE = False
 SAVE_NAME = TITLE
 
 # simulations folders
 father_folder: Path = Path(r"C:\corona\validation\matrix-vs-vector")  # optional
-simulations: List[Simulation] = [Simulation(path=father_folder / "no-policy-matrix", prefix="matrix only"),
-                          Simulation(path=father_folder / "no-policy-vector", prefix="vector only")]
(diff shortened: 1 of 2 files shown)


analysis-tools/multiple_runs_comparison.py
```diff
@@ -0,0 +1,57 @@
...
+                   if
+                   (os.path.isdir(self.path / folder_name) and not folder_name == "matrix_analysis")]
+
+        self.results = []
+        for folder in folders:
+            self.results.append(pd.read_csv(folder / "final_results.csv"))
+
+    def plot_given_stats(self, stats_to_plot: List[str], ax: axis):
+        for stat in stats_to_plot:
+            means = np.mean([result[stat] for result in self.results], axis=0)
```
bgmoshe: Please also add median

analysis-tools/multiple_runs_comparison.py
```diff
@@ -0,0 +1,57 @@
...
+
+# configuration:
+STATS_TO_COMPARE = ["was ever sick"]
+TITLE = "matrix vs vector no policy comparison"
+TO_SHOW = False
+TO_SAVE = True
+SAVE_NAME = TITLE
+
+# simulations folders
+father_folder: Path = Path(r"C:\corona\validation\matrix-vs-vector")  # optional
```
bgmoshe: *parent
Did you just assume the folder's gender???

bgmoshe: Why hardcoded? Give as input or put in a json config file

bgmoshe:
@tal-nc  Name-wise, calling it comparing tool is incorrect. It *plots* multiple runs

analysis-tools/multiple_runs_comparison.py
tal-nc: switched to parent.
i think this is small enough to have it's configuration inside of it.
this is suppose to be similar to the Israel comparison script.

bgmoshe (approved):

PR #366, opened 2020-05-05
bugfix

tal-nc:


bgmoshe:
@tal-nc  What is the bug and how did this solve it

idanmiara (approved):
this command allows alias to use in the script i.e. py=

PR #367, opened 2020-05-05
Random connections analyzer

tal-nc:
added a new module to analyze random connections


src/corona_hakab_model/analyzers/random_connections_analysis.py
```diff
@@ -0,0 +1,100 @@
...
+            df = pd.DataFrame(data=data)
+            df.to_csv(self.raw_output_folder / f"raw {con_type.name}.csv", index=False)
+
+    def analyze_connections_histograms(self) -> List[Histogram]:
+        histograms = []
+        for con_type in ConnectionTypes:
+            distribution_of_connections = self.num_of_random_connections[:, con_type]
+            average_conn = np.mean(distribution_of_connections)
+            median_conn = np.median(distribution_of_connections)
+            max_connection = max(distribution_of_connections)
```
meirhalachmi: np.max\min is much faster than python's max\min

meirhalachmi: Although I guess that runtime isn't critical in this script

meirhalachmi (approved):

PR #369, opened 2020-05-05
added connection type to gen and run, added deceased to sick agents e…

ghost:
…xport


meirhalachmi:
Why did you add "deceased" to the list of sick states?

ghost:
> Why did you add "deceased" to the list of sick states?

because we need the amount of deceased agents by age distribution (we also need the same information for icu and need close medical care)

not sure this is the best way. but it was the easiest way

meirhalachmi (approved):

PR #371, opened 2020-05-06
switch other connections to be completely exponential and not scale free

tal-nc:


src/corona_hakab_model/generation/matrix_generator.py
```diff
@@ -159,7 +159,7 @@ def __init__(
                     )
                 # TODO note that this currently does the same as scale free graph
```
meirhalachmi: is this TODO still relevant?

tal-nc: nope, removed

meirhalachmi: :)

meirhalachmi (approved):

PR #376, opened 2020-05-12
Feature/add weighted random and daily contagiousness

erangold:


src/corona_hakab_model/common/util.py
```diff
@@ -25,7 +30,11 @@ def off_binom(a, c, b):
     if len(args) == 1:
         return const_dist(*args)
     if len(args) == 2:
-        return uniform_dist(*args)
+        if type(args[0]) == list and type(args[1]) == list:
```
bgmoshe: do not use "type(...) =="
If you want to do this check use isinstance

src/corona_hakab_model/common/agent.py
```diff
@@ -44,14 +46,32 @@ def set_test_result(self, test_result):
...
 
+    def days_in_current_state(self):
+        return self.manager.current_step - self.state_entry_day
+
+    def update_contagiousness_day_offset(self, offset):
+        self.contagiousness_day_offset = offset
+
+    def update_contagiousness(self):
+        state_contagiousness_vector = self.medical_state.contagiousness[self.age]
+        # self.manager.contagiousness_vector[self.index] = state_contagiousness_vector
```
bgmoshe: Remove commented code

src/corona_hakab_model/common/agent.py
```diff
@@ -44,14 +46,32 @@ def set_test_result(self, test_result):
...
         self.manager.test_willingness_vector[self.index] = new_state.test_willingness
 
+    def days_in_current_state(self):
+        return self.manager.current_step - self.state_entry_day
+
+    def update_contagiousness_day_offset(self, offset):
+        self.contagiousness_day_offset = offset
+
+    def update_contagiousness(self):
+        state_contagiousness_vector = self.medical_state.contagiousness[self.age]
```
bgmoshe, line 65: But this is a flaot, no?

src/corona_hakab_model/common/medical_state.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, *args, test_willingness: float, **kwargs):
         super().__init__(*args, **kwargs)
         self.test_willingness = test_willingness
 
-        self.contagiousness = BucketDict()
+        self.contagiousness = BucketDict({0:[0]})
```
bgmoshe, line 22: Why you update it here? Let the user type in in consts ...

src/corona_hakab_model/common/state_machine.py
```diff
@@ -90,6 +90,13 @@ def transfer(self, agents: Set[Agent], origin_state: State) -> Iterable[PendingT
             agent_dest = dests[agent_age][current_dest_index] # Destination of current agent
             current_duration_index = age_dest_duration_index[(agent_age, agent_dest)]
             time_to_next_state = agents_durations[(agent.age, agent_dest)][current_duration_index]
+            if type(time_to_next_state) is tuple:
```
bgmoshe: isinstance

src/corona_hakab_model/common/util.py
```diff
@@ -13,6 +14,10 @@ def uniform_dist(a, b):
         range_to_choose_from = list(range(a, b+1))
         return partial(lambda size=None: np.random.choice(range_to_choose_from, size=size))
 
+    def weighted_dist(elements, p):
+        return partial(lambda size=None: random.choices(elements, weights=p, k=size))
```
bgmoshe: Why not use numpy? No need for new module...

src/corona_hakab_model/common/util.py
```diff
@@ -25,7 +30,11 @@ def off_binom(a, c, b):
     if len(args) == 1:
         return const_dist(*args)
     if len(args) == 2:
-        return uniform_dist(*args)
+        if type(args[0]) == list and type(args[1]) == list:
+            assert len(args[0]) == len(args[1]), f"Elements and weights vectors should be in same size! (elements: {len(args[0])}, weights: {len(args[1])})"
```
bgmoshe, line 68: What if both are empty?

src/corona_hakab_model/manager.py
```diff
@@ -152,6 +149,9 @@ def step(self):
         medical_machine_step_result = self.medical_state_manager.step(new_infection_cases.keys())
         self.new_sick_counter = medical_machine_step_result['new_sick']
 
+        for agent in self.agents:
+            agent.update_contagiousness()
```
bgmoshe: Too wastefull... you need to update only if the agent is infectiouss

bgmoshe:
@erangold  how did you test this?

src/corona_hakab_model/common/util.py
erangold: added assert for checking that both are non-empty lists

src/corona_hakab_model/common/state_machine.py
erangold: this line was removed in last commit (it belongs to a deprecated feature I thought would be useful)

src/corona_hakab_model/common/agent.py
erangold: Whole method was removed in latest commit (deprecated feature I wrote which I thought would be useful)

src/corona_hakab_model/common/util.py
erangold: fixed

src/corona_hakab_model/manager.py
erangold: fixed (for each agent, checked if the agent's state is instance of contagious state)

src/corona_hakab_model/common/medical_state.py
erangold: This is a default value, where the state is not contagious (this appears only in immune/susceptible states)

src/corona_hakab_model/common/util.py
erangold: changed to np.random.choice in lastest commit. 
Also, it seems that there's a rounding problem in many of the given days' distributions in the excel sheet. So in such cases (i.e. sum of probs. isn't 1), I just normalize the list of probabilities.

src/corona_hakab_model/manager.py
bgmoshe: Or just go oever the ones that are not susptible :D

src/corona_hakab_model/common/util.py
```diff
@@ -1,31 +1,75 @@
...
         #  an offset-binomial whose mean is c and call it a day
         n = b-a
         p = (c-a)/(b-a)
-        return partial(lambda size=None: np.random.binomial(n=n,
-                                                            p=p,
-                                                            size=size) + a)
+        domain = list(range(a, b + 1))
+
+        binom_rv = stats.binom(n, p, loc = a) # Calculates "Binom(n, p) + a"
+        probs = [binom_rv.pmf(day) for day in domain]
```
bgmoshe, line 59: Why not use numpy binomial like before? Scipy is shit

PR #380, opened 2020-05-17
Bugfix/tests candidates selection

meirhalachmi:


src/corona_hakab_model/detection_model/healthcare.py
```diff
@@ -63,17 +63,10 @@ def testing_step(self):
             # Who can to be tested
             can_be_tested = self._get_testable(test_location)
             test_candidates = want_to_be_tested & can_be_tested
-            test_candidates_inds = set(np.flatnonzero(test_candidates))
+            test_candidates_inds = set(np.flatnonzero(test_candidates))  # TODO: This might be slow on large population
```
bgmoshe: How about np.unique then (you can drop the flatnonzero, I think, and maybe use nonzero on the unique returned

bgmoshe: Sorry, might have something wrong, why this might be slow?

bgmoshe: image

meirhalachmi: Removed the todo

bgmoshe (approved):

